PR #2, opened 2019-09-20
Add a p5 playground into our ml playground

bethanyaconnor:


package.json
```diff
@@ -46,9 +46,10 @@
     "react": "~15.4.0",
     "react-bootstrap": "0.30.1",
     "react-dom": "~15.4.0",
-    "webpack": "4.19.1",
+    "react-p5-wrapper": "0.0.4",
```
bethanyaconnor: I needed to use this very old version of this package because any other versions needed Babel 7.

I can also do this without the package if we think this'll be an issue.

src/activities/creatureCreator/Canvas.js
```diff
@@ -0,0 +1,35 @@
...
+      this.setState({
+        notes: "color: " + color.toString()
+      });
+    }
+  }
+
+  render() {
+    return (
+      <div>
+        <button onClick={this.randomColor.bind(this)}>Random Color</button>
```
madelynkasula: you don't have to bind this function if you use an anonymous function like this:

```suggestion
        <button onClick={e => this.randomColor()}>Random Color</button>
```

bethanyaconnor:
We don't have any commit hooks yet so the build was failing due to linter errors.

joshlory:
Can we use the same p5 version as the main repo?

bethanyaconnor:
What version of p5 does the main repo use? I briefly looked but couldn't find it

PR #5, opened 2019-09-26
Generate eyes for creatures

madelynkasula:
Generates 2 randomly-sized, bilaterally-symmetrical eyes and renders them to the canvas when "good" or "bad" is clicked:
Screen Shot 2019-09-26 at 11 03 17 AM


src/activities/creatureCreator/sketches/index.js
```diff
@@ -1,18 +1,33 @@
...
   p5.setup = () => {
-    p5.background(220);
+    p5.createCanvas(CANVAS_WIDTH, CANVAS_HEIGHT);
+    p5.background(CANVAS_BG_COLOR);
   };
 
-  p5.draw = () => {
-    // console.log("draw!");
-  };
+  const draw = () => {
```
madelynkasula: i refactored this so that we only re-draw the canvas when we need to (i.e., methods that require a canvas draw will need to call this method, rather than it being called every frame). we may not want this in the future, but it saves us a ton of draw cycles for now

madelynkasula:
@bethanyaconnor yeah, they're the same right now lol. i figure we may not differentiate eye shape/size, but i'm working on bodies now, which will obviously be different for good/bad types

PR #15, opened 2019-10-01
Add sliders to the fish generator

bethanyaconnor:
Screenshot 2019-10-01 at 1 46 45 PM


bethanyaconnor:
There are a number of improvements that can be made here: better handling of the initial state and separating the color list out are two that come to mind.

PR #16, opened 2019-10-02
Render fish + KNN data

madelynkasula:
All fish look the same right now, but this adds the ability to render multiple fish to the screen when the component is loaded:

Screen Shot 2019-10-02 at 9 56 07 AM


src/utils/colors.js
```diff
@@ -0,0 +1,45 @@
+export const COLORS = [
```
bethanyaconnor: Thanks for separating this out 😬

src/utils/sketches/fish.js
```diff
@@ -95,6 +97,22 @@ module.exports = p5 => {
     draw();
   };
 
+  p5.getKnnData = () => {
```
bethanyaconnor: nit: can we move this up in this file? This actually contains a lot of really important information that we might want to reference in the future and it's buried kind of deeply in here.

madelynkasula: yeah, good call -- i'll move it up to the top of our p5 sketch function

PR #17, opened 2019-10-02
Classify fish using KNN data

madelynkasula:


src/activities/hoc2019/PondCreator.jsx
```diff
@@ -8,6 +8,11 @@ const P5 = require('../../utils/loadP5');
 
 const FISH_COUNT = 9;
 
+export const Classifiers = Object.freeze({
```
bethanyaconnor: really nit: isn't `Classes` a bit word for this than `Classifiers`? I _think_ classifiers are the things that actually do the classifying?

madelynkasula: lol i'm not sure anymore but i like `Classes`!

bethanyaconnor: Lol yeah I'm definitely not sure about any of the terminology...

madelynkasula: maybe i'll do `ClassType` for now so that it isn't ambiguous with the loaded keyword class

PR #25, opened 2019-10-03
Separate P5 canvas into its own component 

bethanyaconnor:
- Avoids the race condition of initializing the sketch before the element 
- Paves the way for defining fish parameters in an object

All feedback welcome on how to separate the pond creator from the fish!

Screenshot 2019-10-03 at 10 03 43 AM

### DIFF
--- a/src/activities/hoc2019/P5Canvas.jsx
+++ b/src/activities/hoc2019/P5Canvas.jsx
@@ -1,0 +1,62 @@
+import React, {PropTypes} from 'react';
+const P5 = require('../../utils/loadP5');
+//import {sketch} from '../../utils/sketches';
+import Row from 'react-bootstrap/lib/Row';
+import Col from 'react-bootstrap/lib/Col';
+import Button from 'react-bootstrap/lib/Button';
+
+export default class P5Canvas extends React.Component {
+  static propTypes = {
+    fish_data: PropTypes.object.isRequired,
+    canvasId: PropTypes.string.isRequired,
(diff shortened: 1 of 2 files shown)


src/activities/hoc2019/P5Canvas.jsx
```diff
@@ -0,0 +1,61 @@
+import React, {PropTypes} from 'react';
+const P5 = require('../../utils/loadP5');
+//import {sketch} from '../../utils/sketches';
+import Col from 'react-bootstrap/lib/Col';
+import Button from 'react-bootstrap/lib/Button';
+
+export default class P5Canvas extends React.Component {
+  static propTypes = {
+    fish_data: PropTypes.object.isRequired,
```
madelynkasula: nit: use camelcase for js variables

src/activities/hoc2019/P5Canvas.jsx
```diff
@@ -0,0 +1,61 @@
...
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {p5: null};
+  }
+
+  componentDidMount() {
+    const p5 = new P5(this.props.sketch, this.props.canvasId);
+    this.setState({p5: p5});
```
madelynkasula: since the key/value have the same name, we can shorten this line to:
```suggestion
    this.setState({p5});
```

bethanyaconnor: Ah I keep forgetting about this!

PR #31, opened 2019-10-08
Add framework for saving and restoring state

bethanyaconnor:
+test!

I went with Jest because I'm most familiar with it from our redaction libraries. It is notably different from what Dance Party uses (tape) but I was having trouble configuring tape with React. Happy to change this later but I did feel like this change needed tests at this point (and I found a _lot_ of issues through tests).

This logic is largely based off of the comments at https://github.com/tensorflow/tfjs/issues/633 with some consultation of the API for a Tensor.

### DIFF
--- a/test/unit/simpleTrainerTest.test.js
+++ b/test/unit/simpleTrainerTest.test.js
@@ -37,15 +37,15 @@ describe('Simple Trainer tests', () => {
     const classifierDatasetString = trainer.getDatasetJSON();
     trainer.clearAll();
 
-    const retrained_trainer = new SimpleTrainer();
-    retrained_trainer.setTopK(3);
-    await retrained_trainer.initializeClassifiersWithoutMobilenet();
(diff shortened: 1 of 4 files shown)


test/unit/simpleTrainerTest.test.js
```diff
@@ -0,0 +1,51 @@
...
+    trainer.addExampleData([-1, -1], 1);
+
+    const result = await trainer.predictFromData([1, 1]);
+    expect(result.predictedClassId).toEqual(0);
+    expect(result.confidencesByClassId[0]).toEqual(1);
+
+    const classifierDatasetString = trainer.getDatasetJSON();
+    trainer.clearAll();
+
+    const retrained_trainer = new SimpleTrainer();
```
madelynkasula: super nit: convert snakecase variables to camelcase

bethanyaconnor: Good catch!

PR #33, opened 2019-10-09
Show spritesheet fish in the training screen!

bethanyaconnor:


src/activities/hoc2019/PondCreator.jsx
```diff
@@ -17,42 +17,31 @@ const Modes = Object.freeze({
...
+      initialized: false,
+      trainingData: this.generateFish('training', FISH_COUNT),
       currentMode: Modes.Training,
       predictionFish: [], // this.generateFish('prediction',1),
       predictions: []
     };
-    trainer.initializeClassifiers();
   }
 
   generateFish = (canvasPrefix, numFish) => {
```
madelynkasula: we can remove the `canvasPrefix` param here now 🎉

src/activities/hoc2019/PondCreator.jsx
```diff
@@ -81,6 +70,9 @@ export default class PondCreator extends React.Component {
   };
 
   render() {
+    if (!this.state.initialized) {
```
madelynkasula: does `initialized` ever get set to `true`?

madelynkasula: nvm, the answer was right in front of me lol

madelynkasula:
this has a lot of the same changes i was making, so i'm just going to pull this branch and add my updates to `<FishGrid/>` here

src/activities/hoc2019/PondCreator.jsx
bethanyaconnor: Yeah I don't know if this is strictly necessary but I want to avoid trying to use the trainer before it's initialized.

bethanyaconnor:
Yeah sorry about the overlap! Thanks for merging them!

src/activities/hoc2019/SpritesheetFish.jsx
```diff
@@ -1,22 +1,43 @@
+/* eslint-disable react/prop-types */
```
bethanyaconnor: @madelynkasula I needed to disable `react/prop-types` linting on this file as we're importing the shape. I'm going to merge this but would like to get your thoughts on this!

PR #34, opened 2019-10-09
Load from local storage if any is stored

bethanyaconnor:
PondCreator defines functions to store to local storage to persist across refreshes and browser sessions. Builds on the work from #31.

Only visible change is that I also added a "Start over" button as refreshing the page doesn't cut it anymore.

### DIFF
--- a/src/activities/hoc2019/PondCreator.jsx
+++ b/src/activities/hoc2019/PondCreator.jsx
@@ -82,6 +82,12 @@ export default class PondCreator extends React.Component {
     localStorage.setItem(SESSION_KEY, this.state.trainer.getDatasetJSON());
   };
 
+  startOver() {
+    this.state.trainer.clearAll();
+    localStorage.removeItem(SESSION_KEY);
+    this.setMode(Modes.Training);
+  }
+
   render() {
     if (!this.state.initialized) {
       return null;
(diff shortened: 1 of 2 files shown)


src/activities/hoc2019/PondCreator.jsx
```diff
@@ -69,6 +71,23 @@ export default class PondCreator extends React.Component {
...
 
+  loadTraining() {
+    const storedTraining = localStorage.getItem(SESSION_KEY);
+    if (storedTraining) {
+      this.state.trainer.loadDatasetJSON(storedTraining);
+    }
+  }
+
+  saveTraining = () => {
+    localStorage.setItem(SESSION_KEY, this.state.trainer.getDatasetJSON());
```
bethanyaconnor: I'm not sure if it's better to use `localStorage` or `sessionStorage`. Would welcome input!

madelynkasula: i think localStorage sounds like the right solution at this point. i like that localStorage persists for longer, but that may end up causing annoying issues where we need to clear localStorage more often for signed out users.

i think for now, let's go with localStorage, and we can easily swap it out once we have the whole flow a little more nailed down

bethanyaconnor: Yeah that was my thoughts too. My thought was that local storage is more likely to have unexpected persisted data but that's a product question and we could add some messaging around it or something.

PR #35, opened 2019-10-09
Change image path to be relative to the repo

bethanyaconnor:
This still works locally for me and should fix https://code-dot-org.github.io/ml-activities/


bethanyaconnor:
As Maddie is out and this fixes the build, I'm going to merge once Travis is done. Assigning reviewers for visibility.

PR #44, opened 2019-10-14
Generate and predict ocean

bethanyaconnor:
This is mostly a refactor PR to move the generation logic into it's own util file. The new functions are built around testability instead of usability, so they can definitely be changed later.

I'm also just randomly generating test data right now. Once I add more KNN data (soon!) I'll write tests that are more deterministic. These are just to make sure that the functions actually run.

### DIFF
--- a/src/activities/hoc2019/PondCreator.jsx
+++ b/src/activities/hoc2019/PondCreator.jsx
@@ -4,7 +4,7 @@ import SimpleTrainer from '../../utils/SimpleTrainer';
 import Training from './Training';
 import Predict from './Predict';
 import PondResult from './PondResult';
-import {generateRandomFish} from './SpritesheetFish';
+import {generateRandomFish} from '../../utils/generateOcean';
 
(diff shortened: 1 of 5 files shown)


breville:
Just curious: are you also able to use this to do some rough timing measurements?  Am curious whether we have any steps in our process that might need to be done asynchronous of frame rendering, and whether they might even need UI spinners etc if they take a significant amount of time.

bethanyaconnor:
Did some quick calculations and `generateOcean` takes ~30 ms to generate 2000 fish so this shouldn't be a problem.

`filterOcean` takes about 2 seconds when training on 2000 fish and 1 second when training on 25 fish.

This isn't very scientific (particularly as I ran these estimates as part of the test as opposed to in the browser) but we largely consistent. By far the bits I'm most worried about performance wise are setting up the trainer and predictions. Those also seem like the most likely points for us to distract the user with some animation to fill the time.

PR #45, opened 2019-10-14
HTML button over canvas test

breville:
Just a quick demonstration of a native HTML button over the canvas.  Seems like an easy way to get full button functionality.

### portrait browser
Screenshot 2019-10-14 15 48 51

### landscape browser
Screenshot 2019-10-14 15 48 40


bethanyaconnor:
Awesome! Looks like a good path forward

breville:
Using vw/vh doesn't feel quite right yet, because it's not clear how our canvas size will relate to the browser viewport size at this point.

PR #46, opened 2019-10-15
Add KNN data for the fish

bethanyaconnor:
This PR:
- Updates KNN data with normalized and correct KNN data for each piece
- Adds a handful of tests that predict based on a specific attribute. I pre-generated the training set and have stored it in a JSON file. I do randomly generate the set to predict on as it is much larger, but this could add flakiness. If it becomes a problem, let me know.

This code can definitely be improved in various ways, but this will allow us to start building out the UI with predictions.


test/unit/generateOcean.test.js
```diff
@@ -19,4 +34,75 @@ describe('Generate ocean test', () => {
...
   });
+
+  test('Can predict red fish when only picking red fish', async () => {
+    const numPredictionFish = 2000;
+    const trainingOcean = trainingFishJson;
+    const trainer = new SimpleTrainer();
+    await trainer.initializeClassifiersWithoutMobilenet();
+    trainingOcean.forEach(fish => {
+      trainer.addExampleData(Array.from(fish.knnData), fish.colorPalette.bodyRgb[0] > 100 ? 1 : 0);
+    });
```
joshlory: 🙌 this is so cool!

PR #47, opened 2019-10-15
Initial training screen implementation

madelynkasula:
**Note:** I removed a bunch of code because the linter was yelling about it being unused, but I cherry-picked the code to another to be used in subsequent PRs 🤷‍♀ 

Implements a basic "training" page:

out

### DIFF
--- a/src/demo/index.jsx
+++ b/src/demo/index.jsx
@@ -5,15 +5,23 @@ import {init as initTraining} from './modes/training';
 import {init as initPredicting} from './modes/predicting';
 import {init as initPond} from './modes/pond';
 import {setState, getState} from './state';
+import {generateRandomFish} from '../activities/hoc2019/SpritesheetFish';
 
 $(document).ready(() => {
   let canvas = document.getElementById('activity-canvas');
   canvas.width = constants.canvasWidth;
   canvas.height = constants.canvasHeight;
 
+  // Generate some fish
+  let fishes = [];
+  for (let i = 0; i < 100; i++) {
+    fishes.push(getRandomFish(i));
+  }
+
   // Set up state
   const initialState = {
-    currentMode: Modes.Training
+    currentMode: Modes.Training,
+    fishData: fishes
   };
   setState(initialState);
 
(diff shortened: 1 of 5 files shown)


breville:
Very awesome progress!  Three more screens like this and some ML integration and we'll have an MVP for this sequence!

I simplified the button-approach for now if you'd like to pick that up: https://github.com/code-dot-org/ml-activities/pull/48.  It would also be fine to just have the buttons showing below the canvas for a first-pass I think.

What do you think about making the fish slide smoothly into view?  I'm torn about whether we should MVP with no smooth animation vs. trying to get that animation in place now.  The advantage of the former is that smooth animation isn't strictly necessary, but the advantage of the latter is that we ensure we have the right amount of state accounted for and don't paint ourselves into a corner with the chosen technique.

src/demo/index.jsx
```diff
@@ -1,43 +1,32 @@
...
 import {setState, getState} from './state';
+import {generateRandomFish} from '../utils/generateOcean';
 
 $(document).ready(() => {
-  let canvas = document.getElementById('activity-canvas');
-  canvas.width = constants.canvasWidth;
-  canvas.height = constants.canvasHeight;
+  // Generate some fish
+  let fishes = [];
+  for (let i = 0; i < 100; i++) {
```
bethanyaconnor: nit: you could import `generateOcean` and use that here. Does exactly this right now but I figure we can add any logic to generate 15% red fish or whatever in that function.

bethanyaconnor: Oh ah you can't do that. Then I will try and add a better helper method. :)

bethanyaconnor:
Looks awesome! One thing: did you increase the size of the canvas? It renders almost too large on my screen now.

My 2c is that we should add smooth animation later but I definitely understand not wanting to shoehorn it in at the end so I don't feel particularly strongly either way.

madelynkasula:
oh, and @bethanyaconnor -- the initial button approach brendan created added some CSS to the canvas to make it take up 100% of the screen width, but once the canvas is in Code Studio, we can shrink it down to the size we want

madelynkasula:
i lied, i'm going to integrate #48 in a separate PR because i'm having a little trouble integrating it with screen resizing

PR #49, opened 2019-10-15
Add training + predictions

madelynkasula:
Now you can click "like" and "dislike" on the training screen to categorize various fish, click "next" to go to the prediction screen, then view predictions on different fish in this format:
Screen Shot 2019-10-15 at 3 15 04 PM


src/demo/renderer.js
```diff
@@ -133,27 +138,81 @@ function drawTrainingUiElements(state) {
...
+  });
+}
+
+function drawPredictingUiElements(state) {
+  const container = uiContainer();
+  const buttons = [
+    {
+      text: 'predict next fish',
+      id: 'predict-button',
+      onClick: () => {
```
breville: Longer term I don't think we should have click handler implementations in the renderer.  We should think like a react app: any user actions should immediately modify something in the model.  The view, which is implemented in the renderer, should simply render the source of truth as represented in the model.

src/demo/renderer.js
```diff
@@ -133,27 +138,81 @@ function drawTrainingUiElements(state) {
...
+
+function predictionText(state, onComplete) {
+  const fishDatum = state.fishData[state.trainingIndex];
+  state.trainer.predictFromData(fishDatum.fish.knnData).then(res => {
+    onComplete(res);
+  });
+}
+
+function drawPondScreen(state) {
+  console.log('pond!');
```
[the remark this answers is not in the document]
madelynkasula: just a placeholder!

src/demo/renderer.js
madelynkasula: yeah, absolutely. i was thinking of refactoring to something like:

- the models (training, predicting, pond) are implementations of a base class
- the current model is stored in state (`state.model`)
- invoking a click handler will then be something like `state.model.onClickNext()`

+ move all state manipulation to the model so the renderer is only be able to view state and render accordingly.

i think this is what you're thinking of as well, but let me know if you have other thoughts/ideas!

breville: Yes, I think that's a good approach! 

I'm not a fan of inheritance from a base class when the base class has some common code which is sometimes overridden, because that can get complicated to reason through.  But it can be useful if just providing a common interface, if it turns out each model has common entry points  

I do think that we should aim to have the entire current application state stored in our common state object.  That should be the source of truth, to the degree that if you save it, reload the app, and just restore that state object, things should be in the exact same state they were in before.

Sounds like we're on the same page, which is great.

(The downside to having click handlers & one-off show/hide actions, in the renderer or elsewhere, is that it's very susceptible to bugs in which we see too many or too few elements at the same time, because state is distributed amongst many components.  React has essentially solved this entire class of bugs, which is very common in older UI frameworks.)

PR #53, opened 2019-10-16
Basic pond screen

madelynkasula:
Implements a basic pond screen and the ability to "start over" (which clears the trainer of all previously-trained data):

out

### DIFF
--- a/src/demo/renderer.js
+++ b/src/demo/renderer.js
@@ -212,7 +212,83 @@ function predictionText(state, onComplete) {
 }
 
 function drawPondScreen(state) {
-  console.log('pond!');
+  if (state.backgroundImg) {
+    renderBackgroundImage(state.ctx, state.backgroundImg);
+    drawPondFish(state);
+    if (!state.uiDrawn) {
+      drawPondUiElements(state);
+      state.uiDrawn = true;
+      setState(state);
+    }
+  } else {
+    loadBackgroundImage().then(backgroundImg => {
+      state = {...state, backgroundImg};
+      setState(state);
+      drawPondScreen(state);
+    });
+  }
+}
+
+function drawPondFish(state) {
+  predictAllFish(state, fishWithConfidence => {
+    fishWithConfidence = _.sortBy(fishWithConfidence, ['confidence']);
+    const pondFish = fishWithConfidence.splice(0, 20);
+
+    pondFish.forEach(fishDatum => {
+      loadFishImages(fishDatum.fish).then(results => {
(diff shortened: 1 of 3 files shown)


src/demo/renderer.js
```diff
@@ -56,27 +56,32 @@ function loadBackgroundImage() {
...
   ctx.drawImage(img, 0, 0, constants.canvasWidth, constants.canvasHeight);
 }
 
+function initTrainingScreen(state) {
+  // We only want to draw the background and UI elements once,
+  // so render here instead of in drawTrainingScreen().
+  loadBackgroundImage().then(backgroundImg => {
+    renderBackgroundImage(backgroundImg);
+    drawTrainingScreen(state);
+    drawTrainingUiElements(state);
```
breville: These are more examples of where we should just set some state back in our main app, and then let the renderer respond to that state change when it next detects that it's happened.

src/demo/renderer.js
```diff
@@ -211,8 +220,78 @@ function predictionText(state, onComplete) {
...
+          FISH_CANVAS_HEIGHT / 4,
+          state.canvas.height - FISH_CANVAS_HEIGHT / 4
+        );
+        drawFish(fishDatum.fish, results, state.ctx, randomX, randomY);
+      });
+    });
+  });
+}
+
+function predictAllFish(state, onComplete) {
```
breville: We should be doing this in a model, rather than in the renderer, shouldn't we?

src/demo/state.js
```diff
@@ -16,10 +16,11 @@ export const getState = function() {
...
 export const setState = function(newState) {
   if (newState.currentMode) {
+    // When mode changes, reset backgroundImg and uiDrawn.
     state = {
-      ...initialState,
-      fishData: state.fishData,
-      trainer: state.trainer,
+      ...state,
+      backgroundImg: null,
+      uiDrawn: false,
```
breville: It seemed neat that the state functions were totally agnostic about our state.  Do they now need to know about app-specific options, or is that something the app code and perhaps the models can take care of?

madelynkasula: woops, i had this as a WIP and meant to remove it since i didn't end up using it. removing now!

src/demo/renderer.js
madelynkasula: good call, i'll get the model started and move this into it

madelynkasula: okay, i lied, i'm going to do this in a separate PR so this PR doesn't become enormous

PR #62, opened 2019-10-18
Refactor renderer logic into models

madelynkasula:
Start refactoring model logic out of `renderer.js` and into model files (in `models/`)


madelynkasula:
@breville this is still a WIP, but feel free to take a look at the changes to `renderer.js` and `models/train.js` and let me know if you have any feedback! (i plan on taking the same approach for the predict/pond models)

src/demo/renderer.js
```diff
@@ -1,91 +1,37 @@
...
-      break;
-    case Modes.Pond:
-      initPondScreen(state);
-      break;
-    default:
-      console.error('Unrecognized mode passed to renderer init method.');
+export const drawBackground = imgPath => {
+  const canvas = getState().backgroundCanvas;
+  if (!canvas) {
+    return;
```
breville: Since we only call drawBackground on init, if this failed we wouldn't get another chance to draw it for a given mode.  But hopefully we never expect to land in any of these early exit conditions?

breville:
Actually, there's a bit more that we could do.  At the moment, the models know about the renderer, and tell it when to do things, and what specifically to draw (such as the background image filename).  

In an ideal world, the models don't even know there is a renderer.  They are still capable of running the entire simulation, but in a "headless" fashion.  This means they can be unit tested independently of any renderer as well.

The renderer then runs at 60 frames per second, and simply requests the application state, and renders each frame afresh from that.  We could theoretically swap out a 3d renderer for a 2d renderer, and the simulation encapsulated in the models would be none-the-wiser.

I feel like that's the ideal architecture that makes it truly a "reactive" rendering pipeline.

src/demo/renderer.js
madelynkasula: yeah, i did that to match the current experience, but would like to refactor to something retryable (e.g., we set something like `state.backgroundDirty` and the renderer sets the background if `state.backgroundDirty`)

madelynkasula:
> Actually, there's a bit more that we could do. At the moment, the models know about the renderer, and tell it when to do things, and what specifically to draw (such as the background image filename).

i totally agree, and i think my next step will be to have the renderer run a `render` method every frame and re-render any dirty state. this will be necessary for animations, so i think it makes sense as a next step toward making the renderer+models more agnostic.

PR #63, opened 2019-10-18
Create a cache of canvases

bethanyaconnor:
This currently uses a _super_ simple cache of canvases that just reuses the least recently used canvas. I struggled a bit with what the interface with this should be.

`getCanvas` takes in a key that is the cache key and returns the canvas as well as whether there was a cache hit. If there was, it means the caller can assume that the state of the cache is the same as the least time it was used. If it was a miss, the caller can't assume anything about the canvas. Currently `canvasCache` clears the canvas but it doesn't have to.

If you want to know if this works though, this cache is a size of 20 and repeated likes/dislikes resulted in a ceiling of 24 canvases in memory (one is the background, not sure what the other 3 are). The code currently checked into master creates 8 canvases on every like/dislike and does not garbage collect them.

### DIFF
--- a/src/demo/canvasCache.js
+++ b/src/demo/canvasCache.js
@@ -1,4 +1,4 @@
-const CACHE_SIZE = 20;
+const CACHE_SIZE = 30;
 
 export default class CanvasCache {
   constructor() {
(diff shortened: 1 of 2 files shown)


src/demo/renderer.js
```diff
@@ -219,11 +222,17 @@ const drawFish = (fish, results, ctx, x = 0, y = 0) => {
   const bodyAnchor = bodyAnchorFromType(body, body.type);
   results = _.orderBy(results, ['fishPart.type']);
 
+  const intermediateCanvas = canvasCache.getCanvas(`interediate-${fish.id}`)[0];
```
madelynkasula:
```suggestion
  const intermediateCanvas = canvasCache.getCanvas(`intermediate-${fish.id}`)[0];
```

src/demo/canvasCache.js
```diff
@@ -0,0 +1,40 @@
...
+   * Takes a cache key and returns an array of [canvas, hit]
+   * canvas is a canvas reserved for the caller.
+   * hit indicates whether there was a cache hit
+   * The caller should not make assumptions about the state of the canvas if
+   * there was not a cache hit.
+   */
+  getCanvas(key) {
+    const canvasObjectIdx = this.canvases.findIndex(elem => elem.key === key);
+    const canvasObject = this.canvases.splice(
+      canvasObjectIdx !== -1 ? canvasObjectIdx : this.canvases.length - 1,
```
breville: Could you comment and/or break up this logic a little to explain what it's doing?  It feels important :)

breville:
I don't see the `pingCanvas`, that you mentioned, in the code.

As discussed before, I don't see an especially strong reason to release canvases.  Once a canvas has been allocated, I think it's good to maintain it in memory for future use, which also maintains a more stable memory footprint.  I think using the least-recently-used canvas (not just least-recently-rendered, but least-recently-drawn) should serve our purposes since we control how many fish appear on the screen at one time.

bethanyaconnor:
Oops sorry forgot to update the description -- it's updated now.

We don't track when fish are off the screen so this tracks the last time the canvas was accessed. This notably means that the caller can't do much in the way of storing the canvas in a variable, so I had to get rid of all the `fish.canvas` references (that's how I got floating eyes).

src/demo/canvasCache.js
bethanyaconnor: Done!

breville: Thanks, the comments help a bunch.

PR #64, opened 2019-10-18
More refactoring

madelynkasula:


src/demo/index.jsx
```diff
@@ -10,13 +10,12 @@ $(document).ready(() => {
   canvas.width = backgroundCanvas.width = constants.canvasWidth;
   canvas.height = backgroundCanvas.height = constants.canvasHeight;
 
-  const state = setState({
+  setState({
     currentMode: Modes.Training,
     canvas,
     backgroundCanvas,
     uiContainer: document.getElementById('ui-container')
```
breville: As we discussed before, it seems we could take the canvases and the UI element out of the state at some point, if we define this state as "everything we need to persist to restore the previous state upon reload".

src/demo/renderer.js
```diff
@@ -218,55 +233,19 @@ const drawFish = (fish, results, ctx, x = 0, y = 0) => {
...
+    window.requestAnimationFrame ||
+    window.webkitRequestAnimationFrame ||
+    window.mozRequestAnimationFrame ||
+    window.oRequestAnimationFrame ||
+    window.msRequestAnimationFrame ||
+    function(/* function */ callback, /* DOMElement */ element) {
+      window.setTimeout(callback, 1000 / 60);
+    }
+  );
+})();
```
breville: Smooth animation here we come!

breville:
This looks great.  It's exciting to see the models really focusing on doing business & simulation, and the renderer taking care of all the presentation concerns.

PR #69, opened 2019-10-21
Initial words mode

breville:
A very simple initial implementation of the words mode.  Only shows on second and subsequent iterations through the sequence, to simulate the likely user experience in the actual level progression.

Screenshot 2019-10-21 10 31 06

Known issues: design looks to be changing; this implementation treats a word as chosen as soon as it's selected from the dropdown; the last-selected option is still showing when returning to this scene.

### DIFF
--- a/src/demo/models/words.js
+++ b/src/demo/models/words.js
@@ -1,0 +1,67 @@
+import $ from 'jquery';
+import 'babel-polyfill';
+import {setState, getState} from '../state';
+import {initModel} from './index';
+import {Modes} from '../constants';
+import {backgroundPathForMode, createButton} from '../helpers';
+import {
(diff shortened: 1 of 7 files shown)


src/demo/models/words.js
```diff
@@ -0,0 +1,43 @@
...
+    selectList.addEventListener('change', onChangeWord, false);
+    results.push(selectList);
+  }
+  return results;
+};
+
+const uiElements = createDropdowns();
+
+export const init = () => {
+  setState({uiElements: uiElements});
```
madelynkasula: super nit: since the key and value have the same name, you can shorthand this to:
```suggestion
  setState({uiElements});
```

PR #71, opened 2019-10-22
Preload all fish part images once

breville:
Previously, the fish renderer would attempt to load fish part images over the network any time it had to draw a new fish.

Now, we preload all known fish part images upfront in a new "loading" mode, then cache these in canvases, and use those whenever we draw a new fish.

Note that tinting is still done at individual fish render time.


src/demo/renderer.js
```diff
@@ -21,8 +20,16 @@ let prevState = {};
 
 let currentModeStartTime = $time();
 
+let fishPartImages = {};
+
+export const initRenderer = () => {
+  return loadAllFishPartImages();
```
madelynkasula: this is my favorite line of code... ever

PR #72, opened 2019-10-22
Begin implementing wireframes

madelynkasula:
[Video of the current flow](https://drive.google.com/file/d/1lPuif519Er50rigQUtNRHeKr3_-c7pgV/view?usp=sharing)

### DIFF
--- a/src/demo/state.js
+++ b/src/demo/state.js
@@ -10,7 +10,7 @@ const initialState = {
   uiContainer: null,
   uiElements: [],
   headerContainer: null,
-  headerText: '',
+  headerElements: [],
   footerContainer: null,
   footerElements: []
 };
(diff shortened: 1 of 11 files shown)


src/demo/state.js
```diff
@@ -8,7 +8,12 @@ const initialState = {
   trainer: null,
   trainingIndex: 0,
   uiContainer: null,
-  uiElements: []
+  uiElements: [],
+  headerContainer: null,
+  headerElements: [],
+  footerContainer: null,
+  footerElements: [],
+  iterationCount: 0
```
breville: Should we unify this default state with the initial state set in `index.jsx`'s `$(document).ready`?

madelynkasula: i don't think so -- so far, i've been adding all default state in `state.js` and just having `index.jsx` define state related to the DOM. then, `index.jsx` can hook up any state that depends on the current document, and nothing else needs to care about the DOM (e.g., i'm trying to avoid any `document.getElementById()`-esque logic everywhere except `index.jsx`)

if that makes sense 😂

PR #74, opened 2019-10-22
Increase canvas cache size to prevent thrashing on the pond screen

bethanyaconnor:


breville:
As discussed, we probably don't need to have separate intermediate canvases in the cache, nor more than one intermediate canvas ever in existence, in fact.

bethanyaconnor:
Pulled out an intermediate canvas. PTAL.

src/demo/renderer.js
```diff
@@ -27,6 +27,10 @@ export const initRenderer = () => {
 };
 
 const canvasCache = new CanvasCache();
+const intermediateCanvas = document.createElement('canvas');
```
breville: Should we maybe put this inside `initRenderer`?  I encountered an issue developing last night where these module-level variables were being instantiated before we'd really initialized anything else in our app because `$(document).ready` hadn't yet fired, and it was a bit confusing.  It feels like we should hold off on module-specific business logic until our app has gotten up and running...

bethanyaconnor: Yeah I was torn about that but that makes sense. I moved the `intermediateCanvas`, `intermediateCtx, and `cacheCanvas` instantiation into `initRenderer`.

breville: Thanks.  @madelynkasula, this discussion might be of interest in general too.

madelynkasula: ah, good points. let's talk about this in standup today

PR #78, opened 2019-10-22
Initial work on a code blocks dropdown in predict mode

breville:
A button in the top-left corner of the screen toggles whether it's shown or not.

Screenshot 2019-10-22 15 24 40


src/demo/models/predict.js
```diff
@@ -61,3 +75,22 @@ const onClickPredict = () => {
...
+      .fadeIn();
+  } else {
+    $('#show-code').show();
+    $('#hide-code').hide();
+    // Fade out the code.  Stop the current and any pending jquery animations.
+    $('#code')
+      .stop(true, true)
+      .fadeOut();
+  }
+};
```
breville: To be clear I do not like writing UI code like this :)  If we have much more of this kind of UI, then we have a great candidate for either React or a React-like system that re-renders UI based on state rather than dynamically adjusting visibility of different elements in an event handler.

PR #88, opened 2019-10-24
fix package.json repository url

cpirich:
* `package.json` had an outdated reference to `dance-party.git` - now fixed!


cpirich:
This change was merged as part of https://github.com/code-dot-org/ml-activities/pull/114

PR #93, opened 2019-10-24
Remove some similar assets and delete some known-to-be unused assets 

bethanyaconnor:
Some unused ones are sticking around in case we want to integrate later. Preloading the assets made it so much easier to find unused ones and test that I didn't delete used assets!


src/utils/fishData.js
```diff
@@ -760,12 +728,6 @@ const fishComponents = {
...
       type: FishBodyPart.TAIL
     },
-    spikey2: {
-      index: 13,
-      src: 'images/fish/tailFin/Tail_Fin_Spikey2.png',
-      knnData: [],
-      type: FishBodyPart.TAIL
-    },
     square1: {
       index: 14,
```
breville: Do we need to clean up `index`es at some point?

bethanyaconnor: Yes. I want to wait on that for right now because I also want to rearrange some of these...

bethanyaconnor: Actually I have a change to remove all the indices so I'm going to merge this PR and send it out.

PR #94, opened 2019-10-24
Make pond text white on black background for visibility

madelynkasula:
Screen Shot 2019-10-24 at 1 07 22 PM


bethanyaconnor:
🎉 this is so much better thanks!!

PR #97, opened 2019-10-24
Some KNN improvements

bethanyaconnor:
- Add a distinct number to each component so that you could try to get all fish with the same eyes eg
- Remove height : width ratio from the bodies as it was skewed because of the narrow fish
- Add a shape attribute to the bodies

### DIFF
--- a/src/utils/fishData.js
+++ b/src/utils/fishData.js
@@ -15,8 +15,16 @@ export const FishBodyPart = Object.freeze({
 
 const MouthExpression = Object.freeze({
   SMILE: 0,
-  FROWN: 1,
-  NEUTRAL: 2
+  NEUTRAL: 1,
+  FROWN: 2
+});
+
+const BodyShape = Object.freeze({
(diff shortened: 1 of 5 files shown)


bethanyaconnor:
@madelynkasula this PR increases the test proportion back up to 0.7. Feel free to lower it again if it gets flaky again.

src/utils/fishData.js
```diff
@@ -33,7 +41,7 @@ const fishComponents = {
       pectoralFinFrontAnchor: [55, 72],
       dorsalFinAnchor: [23, -15],
       tailAnchor: [107, 41],
-      knnData: [1.46, 7120],
+      knnData: [7120, BodyShape.OVAL],
```
breville: What do these numbers mean, e.g. 7120?

bethanyaconnor: Comment on line 33 explains. I forgot to update and am pushing that updated comment now.

PR #98, opened 2019-10-24
Text/style tweaks

madelynkasula:
Lots of little text/style tweaks, but one particular thing to call out:

- Previously, the button on the Pond screen said "Start Over." Now, it says "Continue" on the 0th iteration, and "Complete" on every iteration after that.

### DIFF
--- a/src/demo/models/pond.js
+++ b/src/demo/models/pond.js
@@ -21,15 +21,10 @@ const staticUiElements = [
     text: 'How did A.I. do?'
   })
 ];
-const footerElements = [
+const staticFooterElements = [
   createButton({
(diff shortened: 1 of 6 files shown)


src/demo/models/pond.js
```diff
@@ -69,13 +64,29 @@ const predictAllFish = state => {
   });
 };
 
+const continueText = state => {
+  return state.iterationCount === 0 ? 'Continue' : 'Complete';
```
bethanyaconnor: Should this maybe say something like "Try again!"? It is a bit weird to me that it says "Complete" then takes me back to the word picking screen.

PR #104, opened 2019-10-25
Cleanup unused activities/code

madelynkasula:
Updated the underwater background to one from Mark's mocks:
Screen Shot 2019-10-25 at 4 40 01 PM

...and deleted a bunch of unused code from our original prototypes!


bethanyaconnor:
🎉 

Is there any of this code that @breville might want as he works on incorporating React into the project?

PR #105, opened 2019-10-26
Simplified UI in React

breville:
This PR converts all the HTML-based UI from jQuery to React.  It also simplifies some of the UI as we start to think about driving the progression via a https://studio.code.org script, rather than doing everything in this app.

Some notable changes:

- We use React to render all the UI.  We register a callback with the state manager so that we can re-render the React UI each time state is changed, though most of the time React will do very little since nothing in the UI will have changed.
- The header & footer areas outside the canvas have been removed and now all UI is inside the canvas area.
- We no longer show options to go backwards in the UI, though they can certainly come back later as we finesse the narrative.
- Use `?words=small` or `?words=large` to specify which words list to use.  We will expand the use of parameterisation to specify fish vs. non-fish mode, and maybe other things like how many items we expect the user to train on, as the progression design requires.
- A placeholder background has been added to each scene.  Buttons are now brighter to look clickable.
- The training question looks disabled when the fish are moving.
- The sorting screen now shows the label as the fish pass the robot, and the fish bob up and down.

screen1b
screen2
screen3
screen4
screen5
screen6


src/demo/helpers.js
```diff
@@ -9,6 +9,9 @@ export const backgroundPathForMode = mode => {
     imgName = 'underwater';
   }
 
+  // Temporarily show background for every mode.
```
bethanyaconnor: Why is this necessary?

breville: I wanted to see how UI elements looked on a background other than plain white, since I presume we will have some kind of background for each scene.

PR #106, opened 2019-10-28
Beginning of cleaning the ocean part

bethanyaconnor:
Accessible with the url param `mode=fishvtrash`.

Selecting trash over fish gives:
Screenshot 2019-10-30 at 4 17 27 PM


src/demo/OceanObject.js
```diff
@@ -149,7 +202,7 @@ export class FishOceanObject extends OceanObject {
     return this.colorPalette;
   }
 
-  drawToCanvas(fishCanvas) {
+  drawToCanvas(fishCanvas, generateLogits = true) {
```
breville: Can we have a comment describing `generateLogits`?

src/demo/canvasCache.js
```diff
@@ -39,12 +39,8 @@ export default class CanvasCache {
...
       canvasObject.key = key;
       const ctx = canvasObject.canvas.getContext('2d');
-      ctx.clearRect(
-        0,
-        0,
-        canvasObject.canvas.width,
-        canvasObject.canvas.height
-      );
+      ctx.restore();
+      ctx.save();
```
breville: We should actually explicitly `save` at the beginning of something and do a matching `restore` at the end since these are managing a stack.

test/unit/simpleTrainerTest.test.js
```diff
@@ -23,14 +24,16 @@ describe('Simple Trainer tests', () => {
...
-    const result = await trainer.predictFromData([1, 1]);
+    trainer.addExampleTensor(tf.tensor([1, 1]), 0);
+    trainer.addExampleTensor(tf.tensor([1, 1]), 0);
+    trainer.addExampleTensor(tf.tensor([1, 1]), 0);
+    trainer.addExampleTensor(tf.tensor([-1, -1]), 1);
+    trainer.addExampleTensor(tf.tensor([-1, -1]), 1);
+    trainer.addExampleTensor(tf.tensor([-1, -1]), 1);
+
+
+
```
breville: blank lines?

src/demo/OceanObject.js
```diff
@@ -6,14 +8,20 @@ import {
...
 } from './helpers';
+import {imagePaths} from '../utils/trashImages';
 import _ from 'lodash';
 
 let fishPartImages = {};
+let trashImages = {};
+// Used to tint the fish components
 let intermediateCanvas;
+// Used to draw the object in order to evaluate it when using mobilenet
+let tmpCanvas;
```
breville: Maybe we should name this specifically, like `evaluationCanvas`?

src/demo/OceanObject.js
```diff
@@ -44,20 +53,35 @@ export const loadAllFishPartImages = () => {
...
-      results.forEach(result => {
-        if (fishPartImages[result.data.partIndex] === undefined) {
-          fishPartImages[result.data.partIndex] = {};
-        }
-        fishPartImages[result.data.partIndex][result.data.variationIndex] =
-          result.img;
-      });
-    }
-  );
+export const initMobilenet = () => {
```
madelynkasula: this method should probably go in `SimpleTrainer.js` or something like a `trainerHelpers.js` file

src/demo/OceanObject.js
```diff
@@ -218,6 +271,27 @@ export class FishOceanObject extends OceanObject {
...
+/*
+ * Trash object that uses one of the images from TrashImages
+ *
+ * */
+export class TrashOceanObject extends OceanObject {
+  randomize() {
+    const idx = Math.floor(Math.random() * imagePaths.length);
+    this.image = trashImages[idx];
+  }
+  drawToCanvas(canvas, generateLogits = true) {
```
madelynkasula: super nit: add an empty line between methods

src/demo/index.jsx
```diff
@@ -18,13 +18,15 @@ $(document).ready(() => {
 
   // Temporarily use URL parameter to set some state.
   const smallWordSet = window.location.href.indexOf('words=small') !== -1;
+  const loadTrashImages = window.location.href.indexOf('mode=fishvtrash') !== -1;
```
madelynkasula: the latest `master` has a helper for this:
https://github.com/code-dot-org/ml-activities/blob/41beede2799155ab02fdc61886901665b1060778/src/demo/helpers.js#L73-L76

bethanyaconnor: Thanks! I think I used it correctly!

src/demo/OceanObject.js
bethanyaconnor: I was moving all the mobilenet stuff out of SimpleTrainer to avoid the trainer needing to know any of the state.

I could move it into `trainerHelpers` though.

bethanyaconnor: Actually, it then saves to a variable in this file so that won't be very clean either. I'm going to leave it here for now and we can refactor it out later.

PR #107, opened 2019-10-28
Refactor ocean objects (e.g. fish) into a more OO pattern

bethanyaconnor:
This will allow us to keep the render logic tied to the object itself rather than in the renderer. The renderer will just give a canvas to the object (eg a fish) and the object will render itself onto the canvas.

A good portion of this code is just moved, particularly most of the logic-intensive code.

To see why this is useful, look at https://github.com/code-dot-org/ml-activities/pull/106/ and see how much easier it was to add a new type of ocean object.


src/demo/toMode.js
```diff
@@ -0,0 +1,9 @@
+import {init as initModel} from './models';
```
bethanyaconnor: I had a 5 file circular dependency, so I moved this method out. Definitely open to other thoughts about where we should put this or how we should split up the `helpers` files.

src/demo/OceanObject.js
```diff
@@ -0,0 +1,210 @@
...
+  return newOceanObject;
+};
+
+export class OceanObject {
+  constructor(id) {
+    this.id = id;
+    this.knnData = null;
+    this.result = null;
+  }
+  randomize() {}
```
madelynkasula: should these unimplemented methods throw an error like:

```
raise new Error("Not yet implemented");
```

so that we know if subclasses are calling these methods without their own implementations?

src/demo/renderer.js
```diff
@@ -339,90 +286,20 @@ const drawSingleFish = (fish, fishXPos, fishYPos, ctx) => {
...
+    /*const fishCtx = fishCanvas.getContext('2d');
     renderFishFromParts(
       fish,
       fishCtx,
       constants.fishCanvasWidth / 2,
       constants.fishCanvasHeight / 2
-    );
+    );*/
+    //ctx.translate(constants.fishCanvasWidth, 0);
+    //ctx.scale(-1, 1);
```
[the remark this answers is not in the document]
bethanyaconnor: oh shoot yes!

src/demo/OceanObject.js
bethanyaconnor: ah yes. I meant to go back and do this!

Added `throw 'Not yet implemented!'` errors to both methods.

src/demo/models/pond.js
```diff
@@ -34,15 +38,16 @@ const predictAllFish = state => {
...
+    const x = randomInt(
       0,
       constants.canvasWidth - constants.fishCanvasWidth / 2
     );
-    fish.y = randomInt(
+    const y = randomInt(
       0,
       constants.canvasHeight - constants.fishCanvasHeight / 2
     );
+    fish.fish.setXY({x, y});
```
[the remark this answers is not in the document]
bethanyaconnor: Gonna go back to the trusty `fishes` variable names but yes :)

src/demo/OceanObject.js
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        intermediateCtx.putImageData(imageData, xPos, yPos);
+      }
+      ctx.drawImage(
+        intermediateCanvas,
+        constants.fishCanvasWidth / 2 - intermediateCanvas.width / 2,
+        constants.fishCanvasHeight / 2 - intermediateCanvas.height / 2,
+        constants.fishCanvasWidth,
+        constants.fishCanvasHeight
```
breville: If we draw fish here, do we still need the `drawSingleFish` function in the renderer?

breville: oops, disregard

PR #122, opened 2019-10-30
Exclude some fish bodies in "small" set

madelynkasula:
Refactors the `?words=small` URL param to `?set=small` so that we can restrict both the 1) set of words the user sees, and 2) the types of fish the user sees.

Each fish component can now define any `exclusions` so that we can filter those components out if necessary.

**Note:** Right now, we are filtering the fish components every time we call `generateOceanObject()`. I'm going to do a follow-up refactor so that this isn't necessary.

### DIFF
--- a/src/demo/OceanObject.js
+++ b/src/demo/OceanObject.js
@@ -58,9 +58,9 @@ export const loadAllFishPartImages = () => {
   );
 };
 
-export const generateOceanObject = (allowedClasses, id) => {
+export const generateOceanObject = (allowedClasses, id, dataSet = null) => {
   const idx = Math.floor(Math.random() * allowedClasses.length);
-  const newOceanObject = new allowedClasses[idx](id);
+  const newOceanObject = new allowedClasses[idx](id, dataSet);
   newOceanObject.randomize();
   return newOceanObject;
 };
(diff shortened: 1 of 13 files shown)


src/demo/OceanObject.js
```diff
@@ -58,13 +58,31 @@ export const loadAllFishPartImages = () => {
...
+  if (!dataSet) {
+    return fishData;
+  }
+
+  let filteredCopy = {...fishData};
+  Object.keys(filteredCopy).forEach(key => {
+    filteredCopy[key] = Object.values(filteredCopy[key]).filter(
+      option => !(option.exclusions || []).includes(dataSet)
+    );
+  });
```
breville: It would be very good to unit test this logic now.

src/utils/fishData.js
```diff
@@ -184,7 +187,8 @@ const fishComponents = {
       dorsalFinAnchor: [76, -13],
       tailAnchor: [144, 56],
       knnData: [12419, BodyShape.TRIANGLE],
-      type: FishBodyPart.BODY
+      type: FishBodyPart.BODY,
+      exclusions: [DataSet.Small]
```
bethanyaconnor: So if a dataset is listed here, this component _isn't_ included in that set?

madelynkasula: yeah. it feels a little backwards, but it made more sense to me to have components opt-out rather than have to remember to opt-in

src/demo/OceanObject.js
madelynkasula: good call, adding tests now

breville: Thanks for that.  I don't usually write a lot of tests, and when I do, they're usually to make sure things keep working after we're past a hectic creative phase, but whenever there is a tricky algorithm, I think it's great to use tests to help get it right up front, and to make sure it never regresses.  My 2c :)

src/demo/helpers.js
```diff
@@ -66,3 +67,23 @@ export const randomInt = (min, max) => {
 export const clamp = (value, min, max) => {
   return Math.min(Math.max(value, min), max);
 };
+
+// Given a key, returns the value from the browser's current query params.
+export const queryStrFor = key => {
+  return queryString.parse(location.search)[key];
+};
+
+export const filterFishComponents = (fishComponents, dataSet) => {
```
breville: A comment on what this function does, and maybe a sample input & output, would be quite helpful.

PR #124, opened 2019-10-30
Call filterFishComponents once when ocean is generated

madelynkasula:
Rather calling `filterFishComponents` every time `generateOceanObject` is called, it is now called once in `generateOcean`, then the filtered components are passed in to `generateOceanObject`.


src/demo/OceanObject.js
```diff
@@ -59,12 +58,13 @@ export const loadAllFishPartImages = () => {
...
+  allowedClasses,
+  id,
+  allowedComponents = fishData
+) => {
   const idx = Math.floor(Math.random() * allowedClasses.length);
-  const newOceanObject = new allowedClasses[idx](
-    id,
-    filterFishComponents(fishData, dataSet)
-  );
+  const newOceanObject = new allowedClasses[idx](id, allowedComponents);
```
bethanyaconnor: hmm I'm curious how this will work when we have other types of ocean objects?

bethanyaconnor: Can you update the method definition on line 73 to reflect this change?

madelynkasula: ah, good point -- i'll refactor to make this work for other types of ocean objects. i think line 73 will stay the same for now because only `FishOceanObject` will have components supplied to its constructor

PR #130, opened 2019-11-01
Display counter + speech bubbles

madelynkasula:
Without a speech bubble:

Screen Shot 2019-11-01 at 1 50 00 PM

With a speech bubble:

Screen Shot 2019-11-01 at 1 51 05 PM

After 5 categorizations, the A.I. bot says "Keep training!".
After 40, A.I. says "Great work! You can continue when you're ready.".
(cc @simplycloud re: speech bubble text)


src/demo/ui.jsx
```diff
@@ -37,7 +40,16 @@ const styles = {
...
   button: {
-    cursor: 'pointer'
+    cursor: 'pointer',
+    backgroundColor: colors.white,
+    fontSize: '120%',
+    borderRadius: 8,
+    border: `2px solid ${colors.black}`,
+    minWidth: 160,
+    padding: '16px 30px',
+    ':focus': {
```
breville: whoa!

PR #131, opened 2019-11-01
Ensure more even distribution of bodies, mouths, eyes, and color palettes in fish

bethanyaconnor:
Note: I would recommend viewing this PR with whitespace disabled and split view. It'll make the fact that I copied and pasted some code into a helper function much easier to review. :)

This PR cycles through the bodies, eyes, mouths, and color palettes and periodically randomizes the order. This ensures an even distribution of the variations but does take away from the randomness a bit (as expected).


src/utils/generateOcean.js
```diff
@@ -40,3 +72,13 @@ export const filterOcean = async (ocean, trainer) => {
   await Promise.all(predictionPromises);
   return ocean;
 };
+
+/*
+ * Takes a list and shuffles it in place
+ */
+const shuffleList = list => {
```
breville: You could also use `_.shuffle`...

bethanyaconnor: lol you're right. I'll change that

PR #132, opened 2019-11-01
Support for instructions mode, with multi-page instructions

breville:
This implements support for an "instructions" mode.  The app supports three of them right now:

```
"intro-instructions"
"fishvtrash-instructions"
"short-instructions"
```

In the standalone app you can view them with the usual URL parameter, e.g. `?mode=intro-instructions`

The instructions can be multi-page.  If there are two or more pages, then dots will be shown below the instructions, allowing the user to click between instructions pages.  Each page provides is own heading.
 The "Continue" button is shown throughout, though it doesn't currently do anything.

Screenshot 2019-11-01 14 16 39

Screenshot 2019-11-01 14 16 56

Screenshot 2019-11-01 14 17 11


src/demo/helpers.js
```diff
@@ -105,3 +106,17 @@ export const filterFishComponents = (fishComponents, dataSet) => {
 
   return filteredCopy;
 };
+
+export const getAppMode = () => {
+  const appMode = _.last(queryStrFor('mode').toLowerCase().split('-'));
```
madelynkasula: i pulled your branch to run locally, and it looks like we just need a safeguard here:

if we load a page without `?mode=` set, this line throws an error:
Screen Shot 2019-11-01 at 2 42 16 PM
Screen Shot 2019-11-01 at 2 41 43 PM

src/demo/ui.jsx
```diff
@@ -58,9 +60,31 @@ const styles = {
...
+  instructionsDots: {
+    textAlign: 'center',
+    fontSize: 60,
+    position: 'absolute',
+    bottom: '20%',
+    left: 0,
+    width: '100%'
+  },
+  activeDot: {
+    color: 'white'
```
madelynkasula: nit: use/add colors from `colors.js`

src/demo/ui.jsx
```diff
@@ -203,6 +230,119 @@ class Button extends React.Component {
...
+                  }
+                >
+                  {'\u25CF'}
+                </span>
+              );
+            })}
+          </div>
+        )}
+        <Footer>
+          <Button style={styles.continueButton} onClick={() => {}}>
```
madelynkasula: should we just remove the continue button since it's not doing anything right now? (i don't feel very strongly about this)

breville: Good point.  I think it's most likely we'll have a button here, so I thought it worth leaving in so that we have an idea how all the UI elements will play together.

PR #133, opened 2019-11-01
UI cleanup + add radium

madelynkasula:


src/demo/ui.jsx
```diff
@@ -46,17 +39,14 @@ const styles = {
     backgroundColor: colors.white,
     fontSize: '120%',
     borderRadius: 8,
-    border: `2px solid ${colors.black}`,
     minWidth: 160,
     padding: '16px 30px',
-    ':focus': {
```
madelynkasula: this didn't work when i added it in #130 😢but should work once Radium is actually being used

breville: Doesn't this PR start it being used?

src/demo/ui.jsx
```diff
@@ -432,6 +410,19 @@ class SpeechBubble extends React.Component {
 }
 
 class ActivityIntro extends React.Component {
+  onClickContinue = () => {
+    const state = getState();
+    if (state.loadTrashImages) {
+      setState({
+        currentMode: Modes.TrainingIntro,
```
breville: Should we have all transitions to a new `currentMode` go through our `toMode` helper?  Amongst other things, it seems to init the model.

src/demo/ui.jsx
madelynkasula: i installed Radium, but haven't actually set it up yet. i'll do that in a follow-up PR to keep this one small-ish

PR #138, opened 2019-11-03
Add model quality tests

uponthesun:
No assertions on these tests since they're mainly intended to be run manually when tuning model quality, for now. They should still be run on PRs though so they don't get broken by refactors.

### DIFF
--- a/src/utils/fishData.js
+++ b/src/utils/fishData.js
@@ -915,3 +915,6 @@ export const initFishData = () => {
 };
 
 export const fishData = fishComponents;
+export {
+  MouthExpression,
+}
--- a/src/utils/fishData.js
+++ b/src/utils/fishData.js
@@ -104,7 +104,7 @@ export const FishBodyPart = Object.freeze({
   EYE: 6
 });
 
-const MouthExpression = Object.freeze({
+export const MouthExpression = Object.freeze({
   SMILE: 0,
   NEUTRAL: 1,
   FROWN: 2
@@ -915,6 +915,3 @@ export const initFishData = () => {
 };
 
 export const fishData = fishComponents;
-export {
-  MouthExpression,
-}
--- a/test/unit/qualityTest.test.js
+++ b/test/unit/qualityTest.test.js
@@ -117,6 +117,7 @@ const PartKey = Object.freeze({
 });
 
 const NUM_TRIALS = 1;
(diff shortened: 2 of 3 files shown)


src/utils/fishData.js
```diff
@@ -915,3 +915,6 @@ export const initFishData = () => {
 };
 
 export const fishData = fishComponents;
+export {
```
bethanyaconnor: You can just put `export` in front of the definition of `MouthExpression` on line 107.

test/unit/qualityTest.test.js
```diff
@@ -0,0 +1,214 @@
...
+
+  const confusionMatrix = {
+    truePos: truePos,
+    falsePos: falsePos,
+    falseNeg: falseNeg,
+    trueNeg: trueNeg,
+    // If the model didn't predict "yes" for anything in the test data, precision is NaN.
+    // We can assume there was at least one positive in the test data and call this 0 precision.
+    precision: precision || 0,
+    recall: recall,
```
bethanyaconnor: Is the comment for precision not possible for recall as well?

test/unit/qualityTest.test.js
```diff
@@ -0,0 +1,214 @@
...
+describe('Model quality test', () => {
+  beforeAll(() => {
+    initFishData();
+  });
+
+  test('Round fish quality test', async () => {
+    const roundFishFn = (fish) => fish.body.knnData[1] === 0 ? 1 : 0;
+    const trainSize = TRAIN_SIZE;
+
+    const result = await performTrials({numTrials: NUM_TRIALS, trainSize: trainSize, testSize: 100, labelFn: roundFishFn});
```
bethanyaconnor: nit: should be broken up into multiple lines (there are a few of these in this file)

src/utils/fishData.js
uponthesun: 👍

test/unit/qualityTest.test.js
uponthesun: Not unless we use a small sample, since the denominator for recall is # of actual positives in the sample, so it's almost guaranteed there'll be at least one.

test/unit/qualityTest.test.js
uponthesun: 👍

PR #142, opened 2019-11-04
Simple level progression UI on demo page

cpirich:
* Added basic UI in `ml-activities` demo page so you can run through an 8 level progression (the 3 video levels from the 11 level Code Studio script are not included here)
* Using the `?mode=short` URL syntax is still supported to set the initial level that will load
* We now export a new method call `initAll()`. The Code Studio side can get rid of a lot of code and just call this instead. For now, I'm exporting both the numerous methods and the new initAll(). Once we merge a Code Studio change (which I'm already testing locally), we can get rid of most of the exports from here.

Screen Shot 2019-11-04 at 10 38 06 AM

### DIFF
--- a/src/demo/state.js
+++ b/src/demo/state.js
@@ -28,14 +28,22 @@ export const getState = function() {
 };
 
 export const setState = function(newState) {
-  state = {...state, ...newState};
+  return setStateInternal({...state, ...newState});
+};
+
+export const setInitialState = function(newState) {
+  return setStateInternal({...initialState, ...newState});
+};
+
+function setStateInternal(newState) {
+  state = newState;
 
   if (setStateCallback) {
     setStateCallback();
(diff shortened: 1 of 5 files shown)


src/demo/state.js
```diff
@@ -28,14 +28,22 @@ export const getState = function() {
 };
 
 export const setState = function(newState) {
-  state = {...state, ...newState};
+  return setStateInternal({...state, ...newState});
+};
+
+export const setInitialState = function(newState) {
+  return setStateInternal({...initialState, ...newState});
+};
```
breville: @madelynkasula - @cpirich explained to me that this allows us to reinitialise state multiple times in a single app instance, which allows the locally-served version of ml-activities to switch between app modes without doing a page reload.

PR #144, opened 2019-11-04
Add scanners to predict screen

madelynkasula:
**TLDR;** Updates the predict screen to match the wireframes - [video](https://drive.google.com/file/d/1gg3SXKSZTeAnyZUCVOzlwe6wtbAUBBL1/view?usp=sharing)

**What it does:**
- Replaces the "skip" button with "continue"
- Starts the screen with a speech bubble from AI bot and a "Run AI" button. Upon clicking that button, the fish begin to be sorted and a red/green scanner is displayed for each prediction

**Follow-up Work:**
- Replace AI bot image to match prediction


src/demo/renderer.js
```diff
@@ -248,17 +249,49 @@ const drawMovingFish = state => {
...
+  botY = botY || state.canvas.height / 2 - img.height / 2;
+
+  // Move AI bot above fish parade.
+  if (state.isRunning) {
+    botYDestination = botYDestination || botY - 130;
+
+    const distToDestination = Math.abs(botYDestination - botY);
+    if (distToDestination > 1) {
+      const direction = distToDestination === botYDestination - botY ? 1 : -1;
+      botY += direction * botVelocity;
```
breville: The remainder of the animation system has a principle that it relies on the passage of time rather than the number of updates to determine where an item should be on screen.  While not crucial here, it would be nice if this movement did the same thing.  This would be done by interpolating between the source and destination locations based on time passed.

src/demo/renderer.js
```diff
@@ -52,19 +64,14 @@ export const render = () => {
...
   switch (state.currentMode) {
     case Modes.Training:
       drawFrame(state);
       drawMovingFish(state);
       break;
     case Modes.Predicting:
-      setState({
-        canSkipPredict:
-          $time() >= currentModeStartTime + timeBeforeCanSkipPredict
-      });
```
breville: Do we need to remove this?  In the video I found the instant switch of a button to a new one to be a bit jarring.  Unless somebody objects, I'd quite like to leave that delay in.

breville:
This is really awesome!  The scene continues to come alive.

src/demo/renderer.js
madelynkasula: ah yeah, i had that solution initially and was having trouble getting the movement consistent (definitely a bug somewhere i couldn't find), so i'll leave as-is for now and refactor later

PR #146, opened 2019-11-05
Beginning of creaturevtrash mode

bethanyaconnor:
This PR adds sea creatures as well as two new modes: `creaturevtrash` and `creaturevtrashdemo`. `creaturevtrash` should be almost complete except for some minor copy tweaks. `creaturevtrashdemo` (slides 14-19 in the playtest 2 mocks) is not done yet.


src/demo/OceanObject.js
```diff
@@ -79,6 +80,18 @@ export const loadAllTrashImages = () => {
   });
 };
 
+// Load all of the trash assets and store them
```
breville: nit: comment should be changed?

PR #147, opened 2019-11-05
Pond: Fish are now clickable

breville:
In the pond, the fish are now clickable.  When looking for a fish to click on, we look at the fish drawn topmost first, and we skip over a fish if it's already been clicked.  Clicking somewhere without a fish clears the clicked fish.

When a fish is clicked, it's drawn at full size.  The fish are now otherwise drawn at half size.  A TODO has been added to investigate whether some renderers (particularly older software rasterizers) might have performance problems drawing images scaled every drawImage; we might need to pre-cache the half-size fish for this mode.

When a fish is clicked, we also start showing a little text box, constrained to canvas locations, with a simple string describing the confidence of the classification of this fish.  The heuristic is simple and will probably need to be improved.  For now, we log the actual result data structure to the console, which might be handy for analysis.

Screenshot 2019-11-04 21 31 09


src/demo/renderer.js
```diff
@@ -317,28 +317,69 @@ const drawFrame = state => {
...
   const canvas = getState().canvas;
   const ctx = canvas.getContext('2d');
+
+  const fishBounds = [];
+
   getState().pondFish.forEach(fish => {
+    const pondClickedFish = getState().pondClickedFish;
+    const pondClickedFishUs = pondClickedFish && fish.id === pondClickedFish.id;
+
+    const swayMultipleX = pondClickedFishUs ? 120 : 120;
```
madelynkasula: are both values 120 on purpose?

breville: Great question.  For now, yes.  I experimented with reducing the sway when a fish is selected, but this causes the newly-enlarged fish to jump to a different location to where they just were, because it puts them on an entirely new sway path.  I wanted to leave the code like this until I thought of a better approach, but I'll just remove it for now.

src/demo/ui.jsx
```diff
@@ -626,6 +653,54 @@ class Predict extends React.Component {
 }
 
 class Pond extends React.Component {
+  onPondClick(e) {
+    const state = getState();
+    const clickX = e.nativeEvent.offsetX;
```
madelynkasula: out of curiosity, why do we get the offsets from nativeEvent here?

breville: I found this technique online earlier today but now am not sure where.  It seems to avoid having to do additional offsetting such as in https://stackoverflow.com/a/18053642 or https://stackoverflow.com/a/33063222.  I suspect we'll likely need to do something like the latter to handle a scaled canvas as we firm up this code soon.

src/demo/ui.jsx
```diff
@@ -596,11 +671,46 @@ class Pond extends React.Component {
...
+
+      if (!fish.confidence || !fish.confidence.confidencesByClassId) {
+        confidence = 'Not sure';
+      } else if (fish.confidence.confidencesByClassId[0] > 0.99) {
+        confidence = 'Very sure';
+      } else if (fish.confidence.confidencesByClassId[0] > 0.5) {
+        confidence = 'Fairly sure';
+      } else {
+        confidence = 'Not very sure';
+      }
```
breville: @uponthesun Very interested to know whether these heuristics are decent for now, and what they should be going forward...

PR #150, opened 2019-11-05
Playtest fixes

madelynkasula:
Lots of little style tweaks, but most important thing to call out here is that **I removed the `intro-instructions` state since we are removing that level.**

### DIFF
--- a/src/demo/ui.jsx
+++ b/src/demo/ui.jsx
@@ -20,7 +20,7 @@ const styles = {
     display: 'flex',
     alignItems: 'center',
     justifyContent: 'center',
-    fontSize: 48,
+    fontSize: 28,
     lineHeight: '52px'
   },
   body: {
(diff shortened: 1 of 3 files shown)


src/demo/ui.jsx
```diff
@@ -260,28 +262,18 @@ let Button = class Button extends React.Component {
...
-      ]
-    }
-  ],
-
   fishvtrash: [
     {
-      heading: 'Train A.I. to Clean Ocean',
+      heading: 'Train AI to Clean the Ocean',
       text: [
+        'In the following activity we’ll learn about <b>artificial intelligence</b> (AI) and <b>machine learning</b>.',
```
breville: Will this work on studio.code.org?  Due to our webfont, I think we might need to refer to a specific font-face...

madelynkasula: ohh, you're right. i'll revert that change here and pull it out into its own PR so the rest of this can be merged

PR #153, opened 2019-11-05
Make pond fish count accurate

madelynkasula:
Before, if we found 20+ fish for the pond, the text always said we found 20 fish. This updates the state to capture the actual number of fish we found and is reflected in the text.

### DIFF
--- a/src/demo/ui.jsx
+++ b/src/demo/ui.jsx
@@ -673,7 +673,7 @@ class Pond extends React.Component {
     const state = getState();
     const pondText = [
       `Out of ${state.fishData.length} objects, I identified ${
-        state.pondFish.length
+        state.totalPondFish
       } that are ${state.word.toUpperCase()}.`,
       'How did I do?',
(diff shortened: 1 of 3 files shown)


src/demo/ui.jsx
```diff
@@ -714,15 +714,13 @@ class Pond extends React.Component {
...
-              <div>
-                {text}
-              </div>
-            );
-          })}
-        </div>}
+        {state.canSeePondText && (
+          <div style={styles.pondText}>
+            {pondText.map((text, index) => {
+              return <div key={index}>{text}</div>;
```
madelynkasula: note because this diff looks weird: i added `key={index}` here because React requires it

breville: Same :)
https://github.com/code-dot-org/ml-activities/pull/152/files#diff-0dfa01a8a25303d3430c725b8ffc48ceR722-R728

PR #154, opened 2019-11-05
Updates to the bias demonstration

bethanyaconnor:


src/demo/ui.jsx
```diff
@@ -597,28 +610,40 @@ class Predict extends React.Component {
     const state = getState();
     const speechBubbleText = this.speechBubbleText(state);
 
+    const biasText =
+      'There are lots of creatures in the sea who don’t look like fish.\n\nBut that doesn’t mean they should be removed!';
```
breville: Do these newlines do anything when rendered to HTML?

bethanyaconnor: nope! I might just remove them -- they take up too much space if there is a line.

PR #156, opened 2019-11-05
More pond UI changes

breville:
Screenshot 2019-11-05 11 34 43

### DIFF
--- a/src/demo/models/pond.js
+++ b/src/demo/models/pond.js
@@ -38,9 +38,12 @@ const arrangeFish = fishes => {
       0,
       constants.canvasWidth - constants.fishCanvasWidth / 2
     );
+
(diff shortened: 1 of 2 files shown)


src/demo/models/pond.js
```diff
@@ -40,7 +40,7 @@ const arrangeFish = fishes => {
     );
     const y = randomInt(
       0,
-      constants.canvasHeight - constants.fishCanvasHeight / 2
+      constants.canvasHeight - constants.fishCanvasHeight / 2 - 160
```
madelynkasula: what's the purpose of this change?

breville: Good question.  It stops fish showing behind the UI.  I updated with a variable and an explanation.

PR #174, opened 2019-11-07
New fish assets

bethanyaconnor:
New fish assets! 🎉 🐟 🎉 

Screenshot 2019-11-07 at 2 43 17 PM

Major changes:
- All new assets
- Dorsal fin positioning is now determined differently (the bodies set an anchor point for where the bottom of the dorsal fin should be then the dorsal fin moves around that)
- The pectoral fins are mirrors of each other

Bugs:
- Some fish don't fit into their canvas
- Some tails still aren't being colored completely 🙁 

Other things:
- I still want to fix the faces. Some look great and some look like a hot mess


bin/determineKnnData.js
```diff
@@ -17,7 +17,24 @@ body_image_files.forEach(fileName => {
         numPixels++;
       }
     }
-    console.log(image.src, (1.0 * image.width) / image.height, numPixels);
+    const name = fileName.split('.')[0];
+    const src = `${name}_image`;
+
+    console.log(`import ${src} from '../../${imagePath}'`);
```
breville: Intentional logging?

bethanyaconnor: Yep! This script outputs a bunch of lines to put into `fishData`. It can probably be improved to put them in a more usable order though...

uponthesun: Yea I was just using `grep` to work around that for now, but it would definitely be an improvement.

bethanyaconnor: oh that's smart. I just ran it twice with various lines commented out. It was still super helpful and probably saved me an hour or so of writing boilerplate!

PR #179, opened 2019-11-08
Initial Guide UI

breville:
Initial work on a more modular instruction system that I'm calling the Guide.

After a classroom test during the week, we realised that we wanted to scaffold a lot more of the experience, at a deliberate pace, with instructions taking on the role of a teacher talking the viewer through the tutorial experience.  This first, rough pass at such a system adds 29 hand-crafted "Guides" which shows when certain conditions are met.  It's all very similar to the tutorial section of a game.

fish-guide-ui-1

The text "types out" to slow down the viewer's attention.  It can be modal, dimming the screen, or can float in place until any button is pressed.  Conditions by default simply require the program's state to match all specified values, but can also include a function which is evaluated as well.  Similarly, the text shown by default is static, but can instead be a function which is evaluated to generate the content.  A Guide definition can also specify how and where the Guide is shown, and whether it includes an arrow.
 One Guide shows at a time.

Headings have been removed to avoid competing for attention on the screen.  The final "pond lab" level placeholder has been removed as well.

There are plenty more improvements to come, but this gets the basics in place.


package.json
```diff
@@ -75,5 +75,8 @@
   "bugs": {
     "url": "https://github.com/code-dot-org/ml-activities/issues"
   },
-  "homepage": "https://github.com/code-dot-org/ml-activities#readme"
+  "homepage": "https://github.com/code-dot-org/ml-activities#readme",
+  "dependencies": {
+    "react-typist": "^2.0.5"
```
islemaster: :grin: https://jstejada.github.io/react-typist/

Does this belong in `devDependencies` since we're bundling it?

src/demo/models/guide.js
```diff
@@ -0,0 +1,299 @@
...
+    style: 'BottomMiddleButtons'
+  },
+  {
+    id: 'fishvtrash-training-done',
+    text: "Great work!  You can continue when you're ready.",
+    when: {
+      appMode: AppMode.FishVTrash,
+      currentMode: Modes.Training,
+      fn: state => {
+        return state.yesCount + state.noCount > 15; // TODO: 40
```
islemaster: Seeing that this is mostly something we could serialize to JSON makes me want us to load it dynamically and have a tool for tweaking this script without rebuilding the bundle.

breville: Good point.  It also struck me as a feature that long-term could be hoisted out into data, rather than code, especially if we wanted to build an authoring UI for it.  A nice thing about the current implementation is that our reload times for the app are really fast, and so iteration on this has been very easy.  Though you do lose your place in the UI when the app reloads.

bethanyaconnor: The way this is built should also hopefully make it easier for localization at some point.

PR #186, opened 2019-11-11
Generate color palettes from list of colors

madelynkasula:
Rather than having a list of pre-defined color palettes, we now have a single palette of colors that can be used to generate fish. 

New colors from Mark:

Slice 1 (8)

### DIFF
--- a/src/demo/OceanObject.js
+++ b/src/demo/OceanObject.js
@@ -6,7 +6,8 @@ import {
   bodyAnchorFromType,
   colorForFishPart,
   randomInt,
-  clamp
+  clamp,
+  generateColorPalette
 } from './helpers';
 import {trashImagePaths, seaCreatureImagePaths} from '../utils/imagePaths';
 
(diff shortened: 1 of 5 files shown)


src/demo/helpers.js
```diff
@@ -43,11 +43,6 @@ export const bodyAnchorFromType = (body, type) => {
 
 export const colorForFishPart = (palette, part) => {
   switch (part.type) {
-    case FishBodyPart.MOUTH:
-      // If the mouth should be tinted, return the mouth
-      // color. Otherwise, return null so that it is
-      // not tinted.
-      return part.tinted ? palette.mouthRgb : null;
```
madelynkasula: note on this: Mark said none of the new fish mouth assets should be tinted, so I removed the `tinted` property from `fishData.mouths`

src/demo/OceanObject.js
```diff
@@ -275,10 +275,10 @@ export class FishOceanObject extends OceanObject {
     const xPos = bodyAnchor[0] + anchor[0];
     const yPos = bodyAnchor[1] + anchor[1];
     if (part.type === FishBodyPart.PECTORAL_FIN_BACK) {
-          intermediateCtx.translate(xPos+img.width,yPos);
+      intermediateCtx.translate(xPos + img.width, yPos);
```
bethanyaconnor: I'm surprised the linter never caught this....

PR #189, opened 2019-11-12
UI tweaks

breville:
- Remove instructions
- Better responsive sizing for buttons
- Support singular count in pond sentence


breville:
@bethanyaconnor Good catch!  I forgot we respond independently to vertical height changes.  Mitigation now added.

PR #190, opened 2019-11-12
Small word list changes

bethanyaconnor:
Two columns of words: colors and shapes. The options in each column are randomized.
Screenshot 2019-11-12 at 9 58 58 AM

### DIFF
--- a/src/demo/ui.jsx
+++ b/src/demo/ui.jsx
@@ -565,7 +565,7 @@ class Words extends React.Component {
         }
       });
     }
-    
+
     this.state = {choices};
   }
 
--- a/src/demo/ui.jsx
+++ b/src/demo/ui.jsx
@@ -53,20 +53,11 @@ const styles = {
     backgroundColor: colors.blue,
     color: colors.white
   },
-  words2Columns: {
-    columns: '2 auto',
-    marginLeft: '32%',
-    marginRight: '22%'
-  },
   button2col: {
     width: '20%',
-    marginTop: '2%',
-    marginRight: '6%'
-  },
-  words3Columns: {
-    columns: '3 auto',
-    marginLeft: '6%',
-    //marginRight: '6%'
+    marginLeft: '14%',
+    marginRight: '14%',
+    marginTop: '2%'
   },
   button3col: {
     width: '20%',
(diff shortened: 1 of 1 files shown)


src/demo/ui.jsx
```diff
@@ -543,8 +546,26 @@ class Words extends React.Component {
   constructor(props) {
     super(props);
 
-    // Randomize word choices and set in state.
-    const choices = _.shuffle(wordSet[getState().appMode].choices);
+    // Randomize word choices in each set, merge the sets, and set as state.
+    const appMode = getState().appMode;
+    const appModeWordSet = wordSet[appMode].choices;
+    let choices = [];
+    let maxSize = 0;
```
uponthesun: Do we need to generalize this to the case where there are a different number of entries in each column? If we don't foresee needing it we could just instead validate that they're all equal to simplify.

src/demo/ui.jsx
```diff
@@ -543,8 +546,26 @@ class Words extends React.Component {
...
+    // Each subset represents a different column, so merge the sets
+    for (var i = 0; i < appModeWordSet.length; ++i) {
+      appModeWordSet[i] = _.shuffle(appModeWordSet[i]);
+      if (appModeWordSet[i].length > maxSize) {
+        maxSize = appModeWordSet[i].length;
+      }
+    }
+    for (i = 0; i < maxSize; ++i) {
```
uponthesun: What's the reason for this specific nested loop - does `choices` need to be in a certain order?

bethanyaconnor: Yeah. The colors all need to be in the left column and the shapes in the right.

uponthesun: But the local variable `choices` that you set in state is just a flat list which is mixed between the two, right?

bethanyaconnor: In order to get it to show up correctly it needs to be separated into the two columns. I was trying to figure out the CSS to do this "better" but couldn't maintain the button styles correctly.

src/demo/ui.jsx
bethanyaconnor: Fair enough. I tend to err on the side of handling array out of bounds (and like errors) even if we don't think they'll be necessary. I'll add a comment here.

src/demo/ui.jsx
madelynkasula: the order only matters for display reasons, so i think it's okay that the end result is a flat list

uponthesun: 👍 bethany explained it to me on slack.

PR #192, opened 2019-11-12
Fix responsive layout for words lists

breville:
Just small tweaks for some edge case layouts.  

Made consistent with main repo's changes in https://github.com/code-dot-org/code-dot-org/pull/31845)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -43,14 +43,15 @@
       }
       button {
         font-size: 120%;
+        padding: 16px 30px;
       }
       @media screen and (max-width: 830px) {
-        button {
+        .words-button {
           font-size: 60%;
         }
       }
       @media screen and (max-width: 970px) {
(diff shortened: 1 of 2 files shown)


public/index.html
```diff
@@ -43,17 +43,20 @@
...
+        padding: 16px 30px;
       }
       @media screen and (max-width: 970px) {
-        button {
+        .words-button {
           font-size: 70%;
+          padding: 10px 20px;
         }
       }
+      @media screen and (max-width: 830px) {
```
bethanyaconnor: Out of curiousity, why move this section down?

breville: Oh, just to have the `max-width`s descending, like the `max-height`s do below.

bethanyaconnor: 🤦‍♀

PR #194, opened 2019-11-13
message and guide updates

simplycloud:
updates to the guide prompts, instructional text, and dialog locations throughout the activity. tested locally to ensure that the prompts appear at the right time and location on each screen.

image

### DIFF
--- a/src/demo/models/guide.js
+++ b/src/demo/models/guide.js
@@ -24,7 +24,7 @@ const guides = [
   },
   {
     id: 'fishvtrash-training-init4',
-    text: "To program A.I., use the buttons to label an image as either 'fish' or 'not fish'. This will train A.I. to do it on it's own. Let's get started!",
+    text: "To program A.I., use the buttons to label an image as either 'fish' or 'not fish'. This will train A.I. to do it on its own. Let's get started!",
     when: {appMode: AppMode.FishVTrash, currentMode: Modes.Training},
     style: 'BottomMiddle',
   },
--- a/src/demo/models/guide.js
+++ b/src/demo/models/guide.js
@@ -4,33 +4,33 @@ import {AppMode, Modes} from '../constants';
 const guides = [
   {
     id: 'fishvtrash-training-init',
-    text: "Let's clean up the ocean!",
+    text: "Garbage dumped in the ocean and rivers affects water health and marine life.  In this activity you will program an artificial intelligence (AI) to identify trash.  Let's clean up the ocean!",
     when: {appMode: AppMode.FishVTrash, currentMode: Modes.Training},
-    style: 'TopLeft',
+    style: 'Center',
     arrow: 'none'
   },
   {
     id: 'fishvtrash-training-init2',
-    text: 'Use these two buttons to train A.I.',
+    text: "Let's meet A.I.",
     when: {appMode: AppMode.FishVTrash, currentMode: Modes.Training},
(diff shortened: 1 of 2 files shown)


src/demo/models/guide.js
```diff
@@ -4,53 +4,86 @@ import {AppMode, Modes} from '../constants';
...
-    text: "Great work!  You can continue when you're ready.",
+    id: 'fishvtrash-training-init3',
+    text: 'A.I. does not know if an object is a fish or trash, but it can process different images and identify patterns.',
+    when: {appMode: AppMode.FishVTrash, currentMode: Modes.Training},
+    style: 'BottomRightCenter',
+    arrow: 'none'
+  },
+  {
+    id: 'fishvtrash-training-init4',
+    text: "To program A.I., use the buttons to label an image as either 'fish' or 'not fish'. This will train A.I. to do it on it's own. Let's get started!",
```
breville: nit: suggest using `'` to surround text so that we can have `"` around the words.

src/demo/models/guide.js
```diff
@@ -4,53 +4,86 @@ import {AppMode, Modes} from '../constants';
...
-    style: 'BottomRightNarrow',
-    hideBackground: true
+    style: 'BottomRightCenter',
+    arrow: 'none'
   },
   {
-    id: 'fishvtrash-training-keeptraining',
-    text: 'Nice work.  Keep training!',
+    id: 'fishvtrash-training-pause2',
+    text: 'Access to clean water could reduce global diseases by 10%.  Keep training to help clean the water!',
```
breville: We can add newline support if that would be useful.

src/demo/models/guide.js
```diff
@@ -4,53 +4,86 @@ import {AppMode, Modes} from '../constants';
 const guides = [
   {
     id: 'fishvtrash-training-init',
-    text: "Let's clean up the ocean!",
+    text: "Garbage dumped in the ocean and rivers affects water health and marine life.  In this activity you will program an artificial intelligence (AI) to identify trash.  Let's clean up the ocean!",
```
breville: Should we add image support so that we can illustrate these points?  I can imagine a thumbnail image to the left of the text here...

src/demo/models/guide.js
```diff
@@ -60,8 +93,9 @@ const guides = [
     textFn: state => {
       return `Out of ${state.fishData.length} random objects, A.I. identified ${
         state.totalPondFish
-      } that belong in water.`;
+      } as 'fish'.`;
     },
+    //TODO after doing a 2nd training iteration these messages no longer show
```
breville: The state's `iterationCount` can be used for that.

src/demo/models/guide.js
breville: Also `it's` should be `its`.

src/demo/models/guide.js
```diff
@@ -4,53 +4,86 @@ import {AppMode, Modes} from '../constants';
...
       appMode: AppMode.FishVTrash,
       currentMode: Modes.Training,
       fn: state => {
-        return state.yesCount + state.noCount > 15; // TODO: 40
+        return state.yesCount + state.noCount >= 5;
       }
     },
-    style: 'BottomRightNarrow',
-    hideBackground: true
+    style: 'BottomRightCenter',
```
breville: Center might work better for this one?

PR #197, opened 2019-11-13
Add new colors for fish

madelynkasula:
Updating the colors to match our new palette:

Fish Palette v2

### DIFF
--- a/src/utils/fishData.js
+++ b/src/utils/fishData.js
@@ -1451,20 +1451,36 @@ const fishComponents = {
   // COLOR KNN DATA: [red, green, blue]
   colors: {
     yellow: {
-      rgb: [255, 203, 72],
-      knnData: [255, 203, 72]
+      rgb: [255, 226, 100],
+      knnData: [255, 226, 100]
+    },
+    darkYellow: {
+      rgb: [249, 197, 0],
+      knnData: [249, 197, 0]
     },
     green: {
       rgb: [143, 194, 63],
       knnData: [143, 194, 63]
     },
+    darkGreen: {
+      rgb: [34, 151, 109],
+      knnData: [34, 151, 109]
+    },
     blue: {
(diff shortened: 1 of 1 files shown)


src/utils/fishData.js
```diff
@@ -1487,8 +1523,18 @@ const fishComponents = {
...
+      exclusions: [
+        AppMode.FishVTrash,
+        AppMode.CreaturesVTrashDemo,
+        AppMode.CreaturesVTrash,
+        AppMode.FishShort
+      ]
+    },
+    darkPink: {
+      rgb: [255, 109, 172],
+      knnData: [255, 109, 172],
```
breville: Is it worth just unifying the `rgb` and `knnData` fields for colours in this file?

madelynkasula: yeah, probably. i thought it might come in handy for them to be separate, but it hasn't so far 🤷‍♀

madelynkasula: ah, nevermind. it looks like we rely on everything in `fishData` to define its `knnData`, so i'll leave that

bethanyaconnor: I don't think it should be difficult to change (eg here: https://github.com/code-dot-org/ml-activities/blob/babf882f895cb2e395790317c53d04c09191825c/src/demo/OceanObject.js#L237 though there will be tests to change too). I think it could be worth doing at some point but probably not as part of this PR.

madelynkasula: yeah, i changed it there, but we more-so rely on it in `initFishData`: 
https://github.com/code-dot-org/ml-activities/blob/babf882f895cb2e395790317c53d04c09191825c/src/utils/fishData.js#L1516-L1517

bethanyaconnor: Ah missed that spot

PR #198, opened 2019-11-13
Remove external dependency

bethanyaconnor:
## Description
Moving the mobilenet model into a place where we can host it ourselves!

## Generating the model
The model was generated with the command ~`tensorflowjs_converter --input_format=tf_hub     'https://tfhub.dev/google/imagenet/mobilenet_v1_100_224/classification/3'  /mobilenet/web_model`~ `tensorflowjs_converter --input_format=tf_hub 'https://tfhub.dev/google/imagenet/mobilenet_v1_025_224/classification/3' /mobilenet/web_model125` ([documentation](https://github.com/tensorflow/tfjs/tree/master/tfjs-converter)). Unfortunately, the `mobilenet` constructor that accepts a `modelUrl` [wasn't introduced until version 2.0.x](https://github.com/tensorflow/tfjs-models/tree/master/mobilenet#loading-the-model) so all the Tensorflow versions were bumped.

## Other notes
- `@tensorflow-models/knn-classifier` is pinned at version 1.1.x because the tests couldn't find the module at version 1.2.x (root cause unresolved).
- The API for the KNN classifier returning the predicted class changed, so now it returns a label. Currently, SimpleTrainer is casting that back to an int but we should change that.


src/demo/OceanObject.js
```diff
@@ -64,7 +64,9 @@ export const loadAllFishPartImages = () => {
 };
 
 export const initMobilenet = () => {
-  return mobilenetModule.load(1, 0.25).then(res => (mobilenet = res));
+  return mobilenetModule
+    .load({version: 1, modelUrl: 'models/mobilenet/model.json'})
```
cpirich: Have we thought about how this will work in Code Studio?
We could make this URL a parameter that we pass in at init time (looks like we are chained quite deep in init methods here, so it may be easier to put it in as a state variable)
Or is there a way to load this from a data URL so we can just bundle this in our package (like we do for images)?

cpirich: I guess the `.bin` files above (referenced from within the JSON) are actually harder to account for.

bethanyaconnor: Yeah I was leaning towards making this URL a parameter because I was having trouble bundling these files the way we do for png

bethanyaconnor: But yeah it will likely require a bit of refactor...

cpirich: It looks like we could solve this by implementing an `IOHandler` and passing that in.
https://www.tensorflow.org/js/guide/save_load#loading_models_with_iohandlers

cpirich:
This upgrade requires us to download an extra 13MB in model data. Can we investigate so it doesn't cause a perf regression?

bethanyaconnor:
Oops my fault! I accidentally used the most accurate but most expensive version of mobilenet. I switched to using one of the least accurate but much cheaper versions (we've been using mobilenet on this side of the spectrum before this change so I don't think this will cause an issue) and removed over 14 MB of bin files.

For the curious: https://github.com/tensorflow/models/blob/master/research/slim/nets/mobilenet_v1.md lists the options. e92b442 switches to `MobileNet_v1_0.25_224` but before I used `MobileNet_v1_1.0_224`

PR #199, opened 2019-11-14
Play, pause, rewind, + fast-forward

madelynkasula:
[Video!](https://drive.google.com/file/d/1Naz_qFE5F1vRhSF9tvkunrd4CTTHDvb0/view?usp=sharing)

On the sorting screen, "video" controls will now appear after the user clicks "Run" (**except during AppMode.CreaturesVTrashDemo**). The user can play, pause, fast-forward (1 and 2x), and rewind (1 and 2x).

**Note:** I originally had a 4x time scale for fast-forwardining/rewinding, but removed it because PEAT tests fail at 4x. 2x time scale is in the "caution" zone for PEAT, so I assume that will be our maximum scale.

Below, the user is fast-forwarding at 2x speed:
Screen Shot 2019-11-14 at 1 17 55 PM


breville:
What happens when paused and fast-forward is pressed?  My gut is that we should hide (or at least disable) fast-forward until we are playing. 

Also, is there a way to rewind at regular speed?

Brainstorming...

I did wonder about the following UI:

```<<   <   ||   >   >>```

And we can have options vanish when they're not available.  This way you can go forward and backward at regular speed, or subsequently at accelerated speed, and we have symmetry between the forward and backward cases.

But then I think this might be nicer:

```<   ||   >```

And once you're moving in a direction, its button changes into an acelerated option, e.g.

```<   ||   >>`

This makes pause always available, and always a consistent click for forward or back, while maintaining subsequent clicks giving acceleration. 

Anyway, the answers to the first two questions are probably more important...

File: src/demo/helpers.js [hunk 1]
```diff
@@ -137,3 +144,28 @@ export const generateColorPalette = (colors, bodyIndex = null) => {
     knnData: bodyColor.knnData
   };
 };
+
+export const currentRunTime = state => {
+  let t = 0;
+  if (state.isRunning) {
+    if (!state.lastStartTime) {
+      state = setState({lastStartTime: $time()});
```
breville: Is it odd for a getter function to be setting state?

src/demo/helpers.js
```diff
@@ -137,3 +144,28 @@ export const generateColorPalette = (colors, bodyIndex = null) => {
     knnData: bodyColor.knnData
   };
 };
+
+export const currentRunTime = state => {
```
breville: Some comments for these new functions would be good.

breville:
This works really nicely.  I must admit that I am not familiar with all of the logic now, and I think a generous helping of comments throughout this logic would probably be useful.  It's great to see that the UI just drives the state and the renderer renders from that state, and everything remains framerate independent which is helpful on devices which will occasionally skip frames.

src/demo/helpers.js
madelynkasula: it is! i'll move that into the main renderer method, since it's already doing some check-state-and-set-state-appropriately logic

src/demo/helpers.js
madelynkasula: good call, adding those now

madelynkasula:
@breville i sprinkled in some comments, but let me know if there's anywhere in particular that could use more exaplanation

src/demo/helpers.js
```diff
@@ -137,3 +144,28 @@ export const generateColorPalette = (colors, bodyIndex = null) => {
...
+    if (clampTime && t > state.moveTime) {
+      t = state.moveTime;
+    }
+  }
+
+  return t;
+};
+
+// Sets the necessary state to stop fish movement at any time, t.
+// Pausing is optional, but defaults to true.
```
breville: What does stopping fish movement mean if we aren't pausing?

PR #203, opened 2019-11-15
Add a stamp to items on sorting screen

madelynkasula:
v1 of adding stamps to the sorting screen:

Screen Shot 2019-11-15 at 11 32 11 AM
Screen Shot 2019-11-15 at 11 32 01 AM


src/demo/renderer.js
```diff
@@ -291,12 +316,16 @@ const drawMovingFish = state => {
...
 
+    let drawPrediction = false;
     if (state.currentMode === Modes.Predicting) {
       if (fish.getResult()) {
         const midScreenX =
           constants.canvasWidth / 2 - constants.fishCanvasWidth / 2;
-        if (Math.abs(midScreenX - x) <= 50) {
+        drawPrediction = x >= midScreenX;
+
+        if (drawPrediction && x - midScreenX <= 50) {
```
breville: I wonder if we should hoist `x - midScreenX <= 50` into a variable as well just so we that we can name this additional condition...

src/demo/renderer.js
```diff
@@ -335,6 +365,37 @@ const drawMovingFish = state => {
   }
 };
 
+// Draws a prediction stamp to the canvas for the given classId.
+// Note: This method requires icons to be cached in predictionImages.
+// Call loadAllPredictionImages() before this method.
+const drawPrediction = (ctx, x, y, classId) => {
+  const rectSize = 210;
```
bethanyaconnor: Where did this number come from?

madelynkasula: mark's comps 🤷‍♀

src/demo/renderer.js
madelynkasula: good call, adding that now

PR #204, opened 2019-11-16
Add scales!

bethanyaconnor:
Overall these anchor points were easier than some others but they do need to be tweaked, especially on the smaller bodies. Also might be worse removing a couple of the bigger sets of scales if we have trouble.

Screenshot 2019-11-15 at 4 35 25 PM


src/utils/fishData.js
```diff
@@ -1448,6 +1479,63 @@ const fishComponents = {
       ]
     }
   },
+  scales: {
+    scales2: {
+      src: scales2_image,
+      knnData: [],
```
breville: Will the scales have an impact on the ML?

bethanyaconnor: Very slightly. I've added them in as data so if someone chooses all fish with the same scales it should show them fish with those scales. I don't think it will have a negative impact on part 2 now that we use SVM.

PR #205, opened 2019-11-16
word-bank-iteration

simplycloud:
iterating on the word bank for the free-play part of the activity.

image

image

### DIFF
--- a/src/demo/ui.jsx
+++ b/src/demo/ui.jsx
@@ -391,22 +391,39 @@ const wordSet = {
   long: {
     text: ['Choose a new word to teach A.I.'],
     choices: [
+  //    [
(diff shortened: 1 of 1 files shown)


src/demo/ui.jsx
```diff
@@ -391,22 +391,39 @@ const wordSet = {
   long: {
     text: ['Choose a new word to teach A.I.'],
     choices: [
+  //    [
```
breville: I would just delete content we don't want.  We can also revert if we want to bring them back.

simplycloud: Thanks @breville. I can do that. Just considering these as placeholders. I'll update.

PR #207, opened 2019-11-17
Initial work to get a set of sounds playing in the app

breville:
- Random yes sounds
- Random no sounds
- Random sort-yes sounds
- Random sort-no sounds
- Random other sounds for guide dismiss & button presses

Huge thanks to @cpirich for contributing to this PR: getting assets loading properly; and allowing the standalone app to play sounds using a borrowed copy of the main repo's sound system, while also allowing the app embedded on studio.code.org to use the host's instance of the sound system.


src/demo/renderer.js
```diff
@@ -6,7 +6,7 @@ import {
   backgroundPathForMode,
   finishMovement,
   currentRunTime,
-  $time
+  $time,
```
bethanyaconnor: nit: stray comma?

PR #208, opened 2019-11-18
guide updates for part one

simplycloud:
breaking the updates into two parts; these updates to the instructional guides for part 1a and part 1b of the activity.


src/demo/models/guide.js
```diff
@@ -1,101 +1,108 @@
...
     id: 'fishvtrash-training-init2',
-    text: "Let's meet A.I.",
+    text: `Let's meet A.I.`,
     when: {appMode: AppMode.FishVTrash, currentMode: Modes.Training},
     style: 'TopRight',
   },
   {
     id: 'fishvtrash-training-init3',
-    text: 'A.I. does not know if an object is a fish or trash, but it can process different images and identify patterns.',
+    text: `A.I. is an artifical inteligence and does not know if an object is a fish or trash, but it can process different images and identify patterns.`,
```
breville: typo: intelligence

src/demo/models/guide.js
```diff
@@ -1,101 +1,108 @@
...
     id: 'fishvtrash-training-init4',
-    text: "To program A.I., use the buttons to label an image as either 'fish' or 'not fish'. This will train A.I. to do it on its own. Let's get started!",
+    text: `To program A.I., use the buttons to label an image as either "fish" or "not fish".  This will teach A.I. to do it on its own.  Let's get started!`,
     when: {appMode: AppMode.FishVTrash, currentMode: Modes.Training},
     style: 'BottomMiddle',
   },
   {
     id: 'fishvtrash-training-pause1',
-    text: "Did you know 1 in 3 people do not have access to clean water?  Keep training to help A.I. identify trash in the water!",
+    text: `Did you know an estimated 17 billion pounds of plastic enters into the ocean from land-based sources each year?  Keep training to help A.I. identify what trash looks like.`,
```
breville: should be: "enters the ocean", no "into"

src/demo/models/guide.js
simplycloud: fixed, good catch.

src/demo/models/guide.js

src/demo/models/guide.js
```diff
@@ -1,101 +1,108 @@
...
-    style: 'BottomRightCenter',
-    arrow: 'none'
+    style: 'RightCenter',
+    arrow: 'none',
+    image: trashBottleImage
   },
   {
     id: 'fishvtrash-training-pause2',
-    text: 'Access to clean water could reduce global diseases by 10%.  Keep training to help clean the water!',
+    text: `Marine debris comes in many shapes and sizes ranging from small plastics to glass bottles to rubber tires, and much more.  Keep training to help A.I. learn.`,
```
breville: suggest: comma after "sizes"

"much more" should be "many more" since "shapes and sizes" are plural

simplycloud: ok, good points. fixed!

src/demo/models/guide.js
```diff
@@ -168,7 +170,7 @@ const guides = [
   },
   {
     id: 'creaturesvtrashdemo-predicting-pause2',
-    text: 'But that doesn’t mean they should be removed!',
+    text: `There are lots of marine life in the ocean that don’t look like fish.  But that doesn’t mean they don't belong in the ocean.`,
```
breville: Can "marine life" be plural?  It reads as singular here, so I think it would read more naturally as something like: 

"There is a lot of marine life in the ocean that doesn't look like fish."

or 

"There are a lot of marine animals that don't look like animals."

Not sure about exact grammatical rule though.

src/demo/models/guide.js
```diff
@@ -188,25 +190,64 @@ const guides = [
     style: 'BottomMiddle',
     arrow: 'none'
   },
+  {
+    id: 'creaturesvtrash-predicting-pause4',
+    text: `Let's train A.I. again`,
```
breville: Missing punctuation at end?

src/demo/models/guide.js
simplycloud: Let's go with:  "There is a lot of marine life in the ocean that doesn't look like fish."

src/demo/models/guide.js
```diff
@@ -188,25 +190,64 @@ const guides = [
...
+        return state.yesCount + state.noCount >= 5;
+      }
+    },
+    style: 'RightCenter',
+    arrow: 'none',
+    image: trash6PackImage
+  },
+  {
+    id: 'creaturesvtrash-training-init3',
+    text: `Marine plastic pollution has impacted at least 267 species worldwide, including 86% of all sea turtle species and 43% of all marine mammal species. Your training of A.I. can make a difference in how well A.I. identifies trash. Keep going.`,
```
breville: First impression: the combination of a real fact and our instructions make it sound as though training A.I. can actually "make a difference" in the real world.

src/demo/models/guide.js
```diff
@@ -188,25 +190,64 @@ const guides = [
...
+      }
+    },
+    style: 'RightCenter',
+    arrow: 'none',
+    image: turtleImage
+  },
   {
     id: 'creaturesvtrash-predicting-init',
-    text: "Now let’s see if A.I. does a better job identifying what should be in the ocean and what shouldn’t.",
+    text: `Do you think A.I. will do a better job idenifying what should be in the water? Let's watch.`,
```
breville: typo: identifying

src/demo/models/guide.js
simplycloud: Or, perhaps "There is a lot of marine life in the ocean that do not look like fish."

src/demo/models/guide.js
simplycloud: Fixed

src/demo/models/guide.js
breville: I like the first one.  The second one would need to be "There is a lot of marine life in the ocean that *does* not look like fish" or "There *are* a lot of marine life in the ocean that do not look like fish."

src/demo/models/guide.js
simplycloud: Is that a good/bad/incorrect impression?

src/demo/models/guide.js
simplycloud: ugh, sorry. fixed.

src/demo/models/guide.js
breville: Oh, it felt like an over-promise, or at least kind of a stretch.

src/demo/models/guide.js
simplycloud: Ok, I prefer this one for now: "There is a lot of marine life in the ocean that does not look like fish."

src/demo/models/guide.js
simplycloud: Fixed, per discussion.

PR #210, opened 2019-11-18
Draw polaroid behind training fish

madelynkasula:
Screen Shot 2019-11-18 at 2 17 23 PM

**Note:** The background of the polaroid was lighter in Figma, but the sock wasn't rendering well on that color, so Mark gave me the okay to try a darker background for the polaroid.


src/demo/renderer.js
```diff
@@ -440,23 +463,6 @@ const drawPredictBot = state => {
   ctx.drawImage(botImg, botX, botY);
 };
 
-// Draw frame in the center of the screen.
```
madelynkasula: unrelated clean-up: deleted this method as it wasn't used anywhere

src/demo/renderer.js
```diff
@@ -547,75 +558,49 @@ export const clearCanvas = canvas => {
...
 
   const canvasCtx = getState().canvas.getContext('2d');
   canvasCtx.globalAlpha = amount;
-  canvasCtx.fillStyle = overlayColour;
-  DrawFilledRect(0, 0, constants.canvasWidth, constants.canvasHeight);
+  DrawRect(0, 0, constants.canvasWidth, constants.canvasHeight, overlayColour);
   canvasCtx.globalAlpha = 1;
-}
-
-const drawRoundedFrame = (
```
madelynkasula: unrelated clean-up: deleted this method as it wasn't used anywhere

src/demo/renderer.js
```diff
@@ -352,7 +351,15 @@ const drawMovingFish = state => {
...
 
-    drawSingleFish(fish, x, y, ctx, 1, drawPrediction);
+    drawSingleFish(
+      fish,
+      x,
+      y,
+      ctx,
+      1,
+      drawPrediction,
+      state.currentMode === Modes.Training
```
bethanyaconnor: Can you comment on what this line is signaling? Just something like `// drawPolaroid`

PR #211, opened 2019-11-19
separate assets and include in dist package

cpirich:
* We now skip the base64 data-uri conversion from `url-loader` for large (>8KB) images and instead use `file-loader` to ensure that these assets are copied into an `assets` directory under `dist` and are included in the published package
* We do something similar for the `model.json` and associated `.bin` file so those are no longer packaged. The `loadHandler()` added recently can now be removed
* Did prep work to package audio assets in the same way (mp3, ogg, wav) into `dist/sounds`
* Clean up the files that are included in the published package
* Clean up the exports for `main.js` to only include `initAll()`
* Added a new `assetPath.js` file (exported alongside `main.js` and `demo.js`) that allows us to redirect the asset path to a different location. Verified that Code Studio can use this to ensure that assets are served from `/blockly/media/skins/fish/assets` (similar to the location used by the minecraft levels)

For reference, this is what is now included in our published package:
```
npm notice === Tarball Contents === 
npm notice 2.9kB   package.json                           
npm notice 10.8kB  LICENSE                                
npm notice 1.5kB   README.md                              
npm notice 4.3kB   dist/assetPath.js                      
npm notice 250.3kB dist/assets/images/blue-scanner.png    
npm notice 233.3kB dist/assets/images/green-scanner.png   
npm notice 9.1kB   dist/assets/images/Octopus.png         
npm notice 265.6kB dist/assets/images/red-scanner.png     
npm notice 1.9MB   dist/assets/models/group1-shard1of1.bin
npm notice 50.0kB  dist/assets/models/model.json          
npm notice 9.8MB   dist/main.js                           
npm notice === Tarball Details === 
```

### DIFF
--- a/src/demo/assetPath.js
+++ b/src/demo/assetPath.js
@@ -1,0 +1,5 @@
+global.__ml_activities_asset_public_path__ = './assets/';
+
+export const setAssetPath = (path) => {
+  global.__ml_activities_asset_public_path__  = path;
+};
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -72,7 +72,7 @@ module.exports = {
   },
   performance: {
     assetFilter: function (assetFilename) {
-      return (/^(images|models|sounds)/.test(assetFilename));
+      return (/^assets\//.test(assetFilename));
     },
     maxAssetSize: 300000,
     maxEntrypointSize: 10500000,
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,9 +1,13 @@
+const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+const CopyPlugin = require('copy-webpack-plugin');
+
 module.exports = {
   devtool: 'eval-cheap-module-source-map',
   resolve: {
     extensions: ['.js', '.jsx'],
   },
   entry: {
+    assetPath: './src/demo/assetPath.js',
     main: './src/index.js',
     demo: './src/demo/index.jsx',
   },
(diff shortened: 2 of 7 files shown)


src/demo/assetPath.js
```diff
@@ -0,0 +1,5 @@
+global.__ml_activities_asset_public_path__ = './assets/';
+
+export const setAssetPath = (path) => {
+  global.__ml_activities_asset_public_path__  = path;
```
cpirich: This is the magic that allows us to use a different asset prefix when running under Code Studio. This API will be called by Code Studio before it does a runtime `require('@code-dot-org/ml-activities')`

webpack.config.js
```diff
@@ -30,11 +34,55 @@ module.exports = {
...
-      test: /\.(png|bin)$/,
+      test: /\.png$/,
       loader: "url-loader",
       options: {
-        limit: true, // everything
+        limit: 8192,
+        outputPath: 'assets/images',
+        publicPath: 'images',
+        postTransformPublicPath: (p) => `__ml_activities_asset_public_path__ + ${p}`,
+        name: '[name].[ext]?[contenthash]',
```
cpirich: The use of `[contenthash]` here ensures that the request URLs encoded into the JS code will change in the event that the content of the files change. This is very helpful in case we modify assets without changing filenames and re-deploy.

webpack.config.js
```diff
@@ -30,11 +34,55 @@ module.exports = {
...
       }
     }],
   },
+  performance: {
+    assetFilter: function (assetFilename) {
+      return (/^(images|models|sounds)/.test(assetFilename));
+    },
+    maxAssetSize: 300000,
+    maxEntrypointSize: 10500000,
```
cpirich: Added these values to ensure that we will see warnings only when new things jump in size.

webpack.config.js
```diff
@@ -30,11 +34,55 @@ module.exports = {
...
+  },
+  plugins: [
+    new CleanWebpackPlugin(),
```
cpirich: This ensures that our `dist` directory is clean before we do a new build

webpack.config.js
```diff
@@ -30,11 +34,55 @@ module.exports = {
...
+    new CopyPlugin([{
+      from: 'src/demo/*.bin',
+      to: 'assets/models/',
+      flatten: true,
+    }]),
```
cpirich: This handles copying the `.bin` file for the mobilenet model. Since we don't import the `.bin` file (it is just referenced from within the `model.json`, we need to explicitly copy it)

PR #213, opened 2019-11-19
Play again button to final pond screen

bethanyaconnor:
Screenshot 2019-11-18 at 5 34 52 PM


src/demo/ui.jsx
```diff
@@ -756,6 +756,19 @@ class Pond extends React.Component {
       }
     }
 
+    const nextButtonText =
+      state.appMode === AppMode.FishLong ? 'Play Again' : 'Continue';
+    const nextButtonOnClick = () => {
+      if (state.appMode === AppMode.FishLong) {
+        resetTraining();
+        toMode(Modes.Words);
```
breville: Philosophical question around whether `toMode` should be responsible for `resetTraining` but no strong opinion.

bethanyaconnor: I don't think so because we also have "Train Again" where we definitely don't want to reset training.

PR #214, opened 2019-11-19
Only train SVM trainer when we have at least two data points

bethanyaconnor:
I think there's an infinite loop in SVM when we only have one data point [here](https://github.com/karpathy/svmjs/blob/b75b71289dd81fc909a5b3fb8b1caf20fbe45121/lib/svm.js#L92). We could try to fix it ourselves as we're already on a fork or we could only train with at least two data points (this fix).


bethanyaconnor:
Updated so that SVMTrainer returns a prediction when only one data point was provided, based on a discussion at stand up (mostly for consistency with KNN). Let me know if you have any concerns!

PR #224, opened 2019-11-21
Guide: UI improvements

breville:
A variety of UI improvements to the Guide, getting it closer to the latest comps.  It still supports showing images in the legacy fashion, in the left part of the box, right now.

The Guide now usually shows up in the lower-center:
Screenshot 2019-11-20 20 45 39

There is optional arrow support:
Screenshot 2019-11-20 20 45 48

Info boxes have a different look:
Screenshot 2019-11-20 20 46 19


madelynkasula:
i'm getting some overflow on the info boxes:
Screen Shot 2019-11-20 at 10 00 34 PM

breville:
@madelynkasula Great catch.  Now fixed.  I also plan on doing another pass to responsively reduce font & button sizes so that this UI looks better on smaller widths.

markabarrett:
The text box and arrows are looking very good, my only wish would be that the text box could stay one height and width, even if those dimensions scaled to fit different devices. For the  'Did You Know' cards- instead of characters you want to go with the garbage, is this your idea or a request from someone?  This could get weird with different sizes of trash object impacting the text in unpredictable ways. Whether we do it this way or separate the art from the text, could we make the card large enough to cover the words above it (e.g. "Is this a fish?") and darken the tint over everything behind it enough that it is very clear that the buttons below are not part of this card. We also need to add the button at the bottom of this card.

breville:
@markabarrett Great feedback; hope to incorporate in next round of refinement.  

The images were chosen by @simplycloud and might be subject to revision.

PR #226, opened 2019-11-21
Pond: Initial explain UI

breville:
This PR adds a new "explain" UI to the pond in the final two levels of the Oceans tutorial.  It presents the SVM explanation information provided in https://github.com/code-dot-org/ml-activities/pull/220.

Upon entering the pond, so long as the user trained on both "likes" and "dislikes", a new information button is shown:

Screenshot 2019-11-23 11 29 51

Pressing it reveals a panel which shows the top four fish parts that were most important in doing this pond's categorisation.  The bar lengths are normalised to the length of the first bar, which is always the longest.

Screenshot 2019-11-23 11 29 56

Selecting an individual fish changes the panel to show the top four fish parts that the system used to determine whether it was in the category (green bars) or not in the category (red bars).  For the pond view of the best-matched fish, the green bars will outweigh the red bars, though again note that we're only showing the top four fish parts.  The bar lengths are normalised to the largest single fish part bar length for the current pond view.

Screenshot 2019-11-23 11 30 12

Selecting a fish on the left side of the pond will move the panel to the right:

Screenshot 2019-11-23 11 33 58

The panel works the same when viewing the least-matched fish in the pond, though the red bars will become predominant for a given fish in this case:

Screenshot 2019-11-23 11 30 26

We have nicer comps to which I'd like to match the UI, and I'd also like to factor the rendering code for this to be nicer, but since we should be feature-complete, like, yesterday, I'm proposing getting this version into the build, and hope to cycle back to it.

In the current implementation, the panel is dismissed when it's clicked or when the screen is otherwise clicked where there isn't a fish.  It's straightforward to adjust this behaviour.


breville:
@bethanyaconnor You're not wrong.  The best technique thus far is to click an exposed fish on the side that you want, so that the box flips to the other side, and then to click around the ones you want.  Open to additional ideas though!

bethanyaconnor:
I instinctively wanted to drag the box or something but I think we can play with it a bit next week and see what's feasible.

PR #230, opened 2019-11-21
Reset training button

bethanyaconnor:
Counter and erase training button in top right:
Screenshot 2019-11-21 at 12 39 00 PM

Confirmation modal:
Screenshot 2019-11-21 at 1 37 00 PM

There's a lot of duplication between the guide and the confirmation modal but the guide modal is pretty baked into the guide logic so I'd like to disentangle them in a separate PR / at a different time.


src/demo/ui.jsx
```diff
@@ -90,6 +94,62 @@ const styles = {
...
+    float: 'left',
+    width: '30%'
+  },
+  confirmationDialogRight: {
+    float: 'right',
+    width: '70%'
+  },
+  confirmationHeader: {
+    fontSize: 40,
+    lineHeight: '40px',
```
madelynkasula: super nit: values like this can be shorthanded to `lineHeight: 40` (like line 126). the only time you _can't_ use that shorthand is for something like lines 141 & 142 where multiple values or a percentage are defined. basically, if you're defining a single value in pixels, you can shorthand to an integer

src/demo/helpers.js
```diff
@@ -171,9 +171,11 @@ export const finishMovement = (t, pause = true) => {
   });
 };
 
-export const resetTraining = () => {
+export const resetTraining = state => {
+  const trainer = state.trainer;
+  trainer.clearAll();
   setState({
-    trainer: null,
+    trainer,
```
madelynkasula: do we still need to re-set `state.trainer` here since we're no longer setting it to `null`?

src/demo/ui.jsx
bethanyaconnor: Turns out we always quote the lineHeight values so I fixed all of them.

bethanyaconnor: Actually it looks like using a number is converting it to a multiplier of the font size: https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

madelynkasula: ohhh, interesting -- good catch! TIL 🌈

PR #231, opened 2019-11-22
Fish grid on results screen

madelynkasula:
[Video](https://drive.google.com/file/d/1gBoiIP_hM9AzpJrq61boLQEr-GvEMwFA/view?usp=sharing)

Organizes fish into a grid on the results screen:

Screen Shot 2019-11-21 at 4 54 06 PM

### DIFF
--- a/src/demo/models/pond.js
+++ b/src/demo/models/pond.js
@@ -33,23 +33,43 @@ const predictAllFish = state => {
 };
 
 const arrangeFish = fishes => {
+  let fishArrangement = formatArrangement();
+
   fishes.forEach(fish => {
-    // Pick a random side of the bot UI.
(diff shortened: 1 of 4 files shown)


src/demo/models/pond.js
```diff
@@ -33,23 +35,48 @@ const predictAllFish = state => {
...
+
+      if (!intermediateArr[col]) {
+        intermediateArr[col] = [];
+      }
+
+      intermediateArr[col].push([colIdx, rowIdx]);
+    });
+  });
+
+  // Flatten nested intermediateArr into a 1-dimensional array.
```
bethanyaconnor: Can you elaborate here what the 1-dimensional array represents (e.g. the coordinates in the order they should be filled)?

PR #236, opened 2019-11-22
Shrink main.js from webpack by using externals

cpirich:
* Shrink our external package `main.js` (used within Code Studio) by 34%: from 9885072 bytes to 6546734 bytes by keeping four packages out of our bundle and labeling them as `externals`: `lodash`, `radium`, `react`, and `react-dom`. These are now listed as `peerDependencies` by our package, indicating that we expect anyone linking our package to also link those packages alongside `ml-activities`
* Stop generating `production` (`main.js`) when running `yarn start` - we only do this when running `yarn build` now
* Removed some unreferenced packages from `package.json`: jquery-ui-touch-punch, jquery-ui, react-bootstrap, react-color
* Added `webpack-bundle-analyzer` as a devDependency so we can easily invoke it if we want to understand our bundle (and generate pretty pictures like the ones below)

Before:
Screen Shot 2019-11-22 at 3 08 23 PM

After:
Screen Shot 2019-11-22 at 3 10 17 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,12 @@
   },
   "author": "",
   "license": "Apache-2.0",
+  "peerDependencies": {
+    "lodash": "^4.17.5",
+    "radium": "^0.25.2",
+    "react": "~15.4.0",
+    "react-dom": "~15.4.0"
+  },
   "devDependencies": {
     "@tensorflow-models/knn-classifier": "~1.1.0",
     "@tensorflow-models/mobilenet": "^2.0.4",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -26,6 +26,12 @@
   },
   "author": "",
   "license": "Apache-2.0",
+  "peerDependencies": {
+    "lodash": "^4.17.5",
+    "radium": "^0.25.2",
+    "react": "~15.4.0",
+    "react-dom": "~15.4.0"
```
davidsbailey: Heads up @jmkulwik that if you bump the react version to 15.6 in our main repo, that you may need to bump these too.

PR #237, opened 2019-11-23
Center the eyes and mouths with each other

bethanyaconnor:
Instead of having independent eye and mouth anchors, have one anchor for the face and place both the mouth and the eyes. In addition, center them based off each other so they look more like faces.

I've talked about doing some scaling on the faces. Still interested in exploring that but that is not included in this PR.

centeredFaces

### DIFF
--- a/src/demo/OceanObject.js
+++ b/src/demo/OceanObject.js
@@ -260,6 +260,27 @@ export class FishOceanObject extends OceanObject {
     return this.colorPalette;
   }
 
+  drawFishFace(bodyAnchor, ctx) {
+    const anchor = this.body.faceAnchor;
+
+    const eyeImg = fishPartImages[this.eye.type][this.eye.index];
+    const mouthImg = fishPartImages[this.mouth.type][this.mouth.index];
+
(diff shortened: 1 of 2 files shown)


src/demo/OceanObject.js
```diff
@@ -351,8 +372,9 @@ export class FishOceanObject extends OceanObject {
     this.drawFishComponent(this.body, bodyAnchor, ctx);
     this.drawFishComponent(this.scales, bodyAnchor, ctx);
     this.drawFishComponent(this.pectoralFinFront, bodyAnchor, ctx);
-    this.drawFishComponent(this.mouth, bodyAnchor, ctx);
-    this.drawFishComponent(this.eye, bodyAnchor, ctx);
+    //this.drawFishComponent(this.mouth, bodyAnchor, ctx);
```
breville: Usual comment: can we get a comment on why this is commented-out for future reference, or just remove it instead?

bethanyaconnor: I'll remove these 😬

PR #239, opened 2019-11-23
Increase set size to fix flakiness

bethanyaconnor:
Most of this diff is due to Prettier fixes.


breville:
How does increasing set size fix flakiness?

bethanyaconnor:
I'm honestly not completely sure it will but this test selects a few fish out of the set of 100. When I added the new body parts a couple days ago, it decreased the number of fish in the set of 100 that matched a certain criteria (i.e. was a circle). I'm pretty sure this test became flaky when I added those new bodies so I'm just trying to increase the set size slightly to see if that fixes it.

bethanyaconnor:
The build still seems to be broken so this might not have fixed it (I of course ran it a few times but it's a bit tricky to evaluate flakiness on this test as it's pretty slow).

PR #242, opened 2019-11-24
Pond: Draw the clicked fish on top of the unclicked fish

breville:
## before

This could happen:

Screenshot 2019-11-23 23 02 00

## after

The same pair of fish but with clicked drawn on top:

Screenshot 2019-11-23 23 06 24

Screenshot 2019-11-23 23 06 17

### DIFF
--- a/src/demo/renderer.js
+++ b/src/demo/renderer.js
@@ -500,30 +500,36 @@ const drawPondFishImages = () => {
   const fishes = state.showRecallFish ? state.recallFish : state.pondFish;
   const fishBounds = [];
 
-  fishes.forEach(fish => {
-    const pondClickedFish = getState().pondClickedFish;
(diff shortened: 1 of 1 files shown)


src/demo/renderer.js
```diff
@@ -500,30 +500,36 @@ const drawPondFishImages = () => {
...
+        const fishBound = drawSingleFish(fish, finalX, finalY, ctx, size);
+
+        // Record this screen location so that we can separately check for clicks on it.
+        fishBounds.push({
+          fishId: fish.id,
+          ...fishBound
+        });
+      }
     });
-    setState({pondFishBounds: fishBounds}, {skipCallback: true});
```
breville: Oops, when I decided React couldn't handle being updated at 60fps, I was probably blasting it a lot more than that.

PR #244, opened 2019-11-24
update to word list

simplycloud:
this change updates the latest words for part 2b of the activity. changes tested locally.

image

image


simplycloud:
confirmed with brendan ok to merge.

PR #245, opened 2019-11-24
updates to guide flow and text

simplycloud:
updates to the instructions and guides include simplified text, some new strings, and changes to how we explain the AI results screen. still a few pending TODO items around new UI elements such as the Info button on the AI results screen. tested locally. also did some clean up to naming and removed imports of images no longer used in guides.

### DIFF
--- a/src/demo/models/guide.js
+++ b/src/demo/models/guide.js
@@ -110,7 +110,7 @@ const guides = [
   },
   {
     id: 'fishvtrash-pond-init1',
-    text: `Based on your training here are the objectes that A.I. identified as "Fish".  How did A.I. do?`,
+    text: `Based on your training here are the objects that A.I. identified as "Fish".  How did A.I. do?`,
     //TODO after doing a 2nd training iteration these messages no longer show
     when: {
       appMode: AppMode.FishVTrash,
@@ -266,7 +266,7 @@ const guides = [
   },
   {
     id: 'creaturesvtrash-pond-init1',
-    text: `Based on your training here are the objectes that A.I. identified as "Belongs in water".  How did A.I. do?`,
+    text: `Based on your training here are the objects that A.I. identified as "Belongs in water".  How did A.I. do?`,
     when: {
       appMode: AppMode.CreaturesVTrash,
       currentMode: Modes.Pond,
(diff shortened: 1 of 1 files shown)


src/demo/models/guide.js
```diff
@@ -95,13 +92,13 @@ const guides = [
     }
   },
   {
-    id: 'fishvtrash-predicting-init',
-    text: `Now let's see if A.I. knows what a "fish" looks like.`,
+    id: 'fishvtrash-predicting-init1',
+    text: `Now let's see if A.I. knows what a "Fish" looks like.`,
```
breville: nit: We've used different double quotes above.. we should be consistent, though it doesn't matter too much which we choose.

src/demo/models/guide.js
```diff
@@ -112,12 +109,8 @@ const guides = [
...
   },
   {
-    id: 'fishvtrash-pond-init',
-    textFn: state => {
-      return `A.I. identified ${
-        state.totalPondFish
-      } objects that match your training data for "fish".  How did A.I. do?`;
-    },
+    id: 'fishvtrash-pond-init1',
+    text: `Based on your training here are the objectes that A.I. identified as "Fish".  How did A.I. do?`,
```
breville: typo: objects

src/demo/models/guide.js
```diff
@@ -276,17 +260,13 @@ const guides = [
...
   },
   {
-    id: 'creaturesvtrash-pond-init',
-    textFn: state => {
-      return `A.I. identified ${
-        state.totalPondFish
-      } objects that match your training data for "belongs in water".  How did A.I. do?`;
-    },
+    id: 'creaturesvtrash-pond-init1',
+    text: `Based on your training here are the objectes that A.I. identified as "Belongs in water".  How did A.I. do?`,
```
breville: typo: objects :)

src/demo/models/guide.js
simplycloud: fixed.

src/demo/models/guide.js

src/demo/models/guide.js
simplycloud: I just pushed an update to address this (i believe) so that anywhere we show the word that AI was being trained on we surround it with " ". Let me know if that addresses your concern. I also fixed one location where I was using the wrong outer quotes. Thanks.

simplycloud:
confirmed with brendan ok to merge.

PR #248, opened 2019-11-24
Fish on words screen

bethanyaconnor:
The fish swim across the screen (I couldn't figure out how to video capture)

Screenshot 2019-11-24 at 3 21 44 PM


src/demo/renderer.js
```diff
@@ -498,9 +504,75 @@ const drawPredictBot = state => {
...
+  const canvas = getState().canvas;
+  const ctx = canvas.getContext('2d');
+  const state = getState();
+
+  const fishScale = 0.7;
+
+  const t = $time();
+  state.wordFish.forEach(fish => {
+    const swayValue = ((t * 360) / (20 * 1000)) % 360;
+    const swayOffsetY = Math.sin(((swayValue * Math.PI) / 180) * 3) / 20;
```
breville: At some point we might want to factor out a wobble helper function :)

src/demo/renderer.js
```diff
@@ -498,9 +504,75 @@ const drawPredictBot = state => {
...
+      0,
+      Math.floor(constants.canvasHeight / (fishScale * constants.fishCanvasHeight)) - 1
+    );
+    const y = lane * constants.fishCanvasHeight * fishScale;
+    newFish.setXY({x: -constants.fishCanvasWidth * 1.5, y});
+    newFish.faceLeft = lane % 2 === 0 ? true : false;
+    wordFish.push(newFish);
+    setState({fishCount: state.fishCount + 1});
+  }
+  setState({wordFish});
```
breville: Is it worth clearing `wordFish` from the state at the end of the scene, rather than just at its `init`, to reduce memory usage?

src/demo/renderer.js
bethanyaconnor: +1 I did just copy and adjust these from the pond.

src/demo/renderer.js
bethanyaconnor: Do we clear any of the fish lists?

PR #249, opened 2019-11-25
Make fish jump into training bot's head

madelynkasula:
[Video](https://drive.google.com/file/d/1zzE3S--6im3HTEuhLQmvWBt_zoFMfhWI/view?usp=sharing)

There will most likely need to be style (i.e., CSS) tweaks on this, but this is the first pass


src/demo/ui.jsx
```diff
@@ -233,9 +235,20 @@ const styles = {
...
     position: 'absolute',
-    height: '40%',
-    top: '28%',
-    left: '76%'
+    height: '100%',
+    top: 0,
+    left: '73%',
+    display: 'flex',
+    flexDirection: 'column',
+    justifyContent: 'center'
```
breville: We'll have to see how this works with the upcoming responsive work too.

src/demo/renderer.js
```diff
@@ -369,7 +368,14 @@ const drawMovingFish = state => {
     }
 
     const drawPolaroid = state.currentMode === Modes.Training;
-    drawSingleFish(fish, x, y, ctx, 1, drawPrediction, drawPolaroid);
+    let size = 1;
+    if (drawPolaroid && state.isRunning && x > midScreenX) {
+      size = 0.35;
+      // Apply sine wave to y-value to make item jump into AI bot's head.
+      y -= Math.sin((runtime / state.moveTime) * Math.PI) * 200;
```
breville: And they said we'd never use sine waves in real life!

src/demo/ui.jsx
breville: Confirmed that this will need to be updated to work responsively.

PR #251, opened 2019-11-25
Responsive: train bot

breville:
The bot opening its head, added in https://github.com/code-dot-org/ml-activities/pull/249, works better with the new responsive system, added in https://github.com/code-dot-org/ml-activities/pull/246.


breville:
This just helps get things in place, but it can do with some tweaking!

breville:
And thanks for making the tweaks!

PR #254, opened 2019-11-25
Generate example fish

bethanyaconnor:
Fun utility to get a canvas of the 100 things in the ocean on that mode.

Best way to use this would be to use the parameters `/?mode={mode}&download=yes`


bethanyaconnor:
Definitely! Not a rush at all!

I've been using the image I download to create Slack emojis but otherwise this isn't an impactful change.

breville:
That is a wonderful use case!

PR #256, opened 2019-11-25
Update fish on words screen

bethanyaconnor:
Instead of randomly generating fish, always have one per lane. Also switch the background of the words screen to underwater.

I don't love the complexity of this change but it is well contained. Let me know if you have any concerns with merging this.


breville:
Maybe some comments explaining the broad strokes of what's happening?

bethanyaconnor:
Added some comments

PR #258, opened 2019-11-25
UI tweaks

madelynkasula:
Hodgepodge of UI tweaks:
- Change Play Again button to say New Word
- Replace "download" icon with polaroid on counter
- Make run button match Code Studio (add 'play' icon to button)
- Highlight selected word on the Word Pick screens (hover state); also fix issue on Yes / No buttons where focus returns to on-click
- Update A.I. bot images
- Align guide arrow with bot on training screen

I did a pass to make sure this all works responsively, so if you notice anything working improperly, let me know!


src/demo/ui.jsx
```diff
@@ -836,6 +855,7 @@ let Train = class Train extends React.Component {
             }}
             sound={'no'}
           >
+            <FontAwesomeIcon icon={faBan} style={{marginRight: 10}} />
```
breville: Worth hoisting these three to a common style?

madelynkasula: yes!

PR #260, opened 2019-11-25
Remove no-op sort call

uponthesun:
The `sortBy` call doesn't do anything right now, since the `confidence` field is never defined on the fish object. (The fish object does have a confidence value, but it's in `.result.confidencesByClassId`.)

I personally think this behavior is better since it's better for allowing students to investigate misclassified fish. Discussed with Anthony as well - see slack thread in channel.

There's supposed to be no functional change in the code with this change, so if you see one, please let me know.

### DIFF
--- a/src/demo/models/pond.js
+++ b/src/demo/models/pond.js
@@ -5,10 +5,9 @@ import constants, {ClassType, AppMode} from '../constants';
 
 export const init = async () => {
   const state = getState();
-  let fishWithConfidence = await predictAllFish(state);
-  fishWithConfidence = _.sortBy(fishWithConfidence, ['confidence']);
+  let fishWithPredictions = await predictAllFish(state);
   const fishByClassType = _.groupBy(
-    fishWithConfidence,
+    fishWithPredictions,
     fish => fish.getResult().predictedClassId
   );
 


src/demo/models/pond.js
```diff
@@ -5,10 +5,9 @@ import constants, {ClassType, AppMode} from '../constants';
 
 export const init = async () => {
   const state = getState();
-  let fishWithConfidence = await predictAllFish(state);
-  fishWithConfidence = _.sortBy(fishWithConfidence, ['confidence']);
+  let fishWithPredictions = await predictAllFish(state);
   const fishByClassType = _.groupBy(
-    fishWithConfidence,
+    fishWithPredictions,
```
madelynkasula: should we still sort the items by confidence?

uponthesun: My opinion is that there's a benefit in just showing a random sample of the fish, instead of the most confident, since it gives a more accurate representation of how well the model is doing, and more opportunities to see misclassified objects and debug them. I could see an argument that we want to show the top results to make the pond look better, though. Maybe it'd be useful to get Mike's input?

madelynkasula: that's a really good point. i think i agree with you. unless there's a good reason to sort them, it's probably better to leave as-is anyways

bethanyaconnor:
Merging this so that this will make the release

PR #264, opened 2019-11-26
Crush pngs

bethanyaconnor:


bethanyaconnor:
For future reference the way I did this was to cd into the directory to crush and run
```at="@" ; for file in *.png; do curl -X POST -s --form "input=$at$file;type=image/png" https://pngcrush.com/crush > "$file-crushed" ; done && mmv -d "*-crushed" "#1" && git add *.png && git commit -m "Use PNG crush on {directory} images"```

breville:
Looks like we also removed some executable permissions on files, which seems reasonable.

PR #265, opened 2019-11-26
Fix instant ending

breville:
A recent regression saw us ending the final pond immediately when our host had provided a function for doing so.  An easy fix.


src/demo/ui.jsx
```diff
@@ -1335,7 +1335,7 @@ let Pond = class Pond extends React.Component {
                 </Button>
                 <Button
                   style={styles.finishButton}
-                  onClick={state.onContinue()}
+                  onClick={() => state.onContinue()}
```
madelynkasula:
```suggestion
                  onClick={state.onContinue}
```

PR #276, opened 2019-11-26
misc guide updates

simplycloud:
updating various guides to meet fit/finish text requirements. adding new guides to encourage more training in part 2 of activity.

### DIFF
--- a/src/demo/models/guide.js
+++ b/src/demo/models/guide.js
@@ -121,7 +121,7 @@ const guides = [
   },
   {
     id: 'fishvtrash-pond-recall',
-    text: `Click here to switch between objects labeled as “Fish” and “Not Fish”.`,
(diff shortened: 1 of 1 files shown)


src/demo/models/guide.js
```diff
@@ -122,11 +121,10 @@ const guides = [
   },
   {
     id: 'fishvtrash-pond-recall',
-    text: `Switch between “fish” and “not fish” here.`,
+    text: `Click here to switch between objects labeled as “Fish” and “Not Fish”.`,
```
madelynkasula: i can't remember what we decided on here -- should it be "fish" / "not fish" (lowercase) or "Fish" / "Not Fish" (titlecase)

breville: Yes it should be lowercase except we literally refer to the Fish/Not Fish buttons in one place.

PR #281, opened 2019-11-26
Start ids from wherever we last left off

bethanyaconnor:
Theoretically we could have a situation where a fish's id is repeated within a small stretch of fish and could cause a cache collision as we were restarting the ids from 0 each time we generate some fish.

Example (generate 10 at a time with cache size of say 2):
generate some fish with ids [0, 3, 8, 2, 9, 1, 7, 5, 4, 6]
then generate some with ids [8, 6, 0, 1, 7, 5, 3, 2, 4, 9]
The last element of the first set will clash with the second element of the second set and will show the wrong fish due to the canvas cache.

### DIFF
--- a/src/demo/models/predict.js
+++ b/src/demo/models/predict.js
@@ -8,8 +8,8 @@ export const init = () => {
   state.trainer.train();
   let fishData = [];
   if (state.appMode === AppMode.CreaturesVTrashDemo) {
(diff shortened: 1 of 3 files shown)


bethanyaconnor:
I can't think of any but I have seen a good number of bugs with cache clashes. I wish I'd worked a bit harder to come up with cache keys a few weeks ago but I'm nervous to mess with that now...

PR #290, opened 2019-11-27
Pond: panel improvements

breville:
After the latest round of bug bash feedback, some pond panel improvements.

Now, the button in the top-right is the only way to toggle panel mode on and off.  Once toggled on, if a fish was already selected, its panel shows right away.  Clicking on no fish just returns the default panel, until panel mode is turned off.

Another change is that the user can now click through the panel to a fish behind it.  This makes it easier to select fish even when the panel is over them.

### DIFF
--- a/src/demo/ui.jsx
+++ b/src/demo/ui.jsx
@@ -371,7 +371,8 @@ const styles = {
     borderRadius: 10,
     left: '3%',
     top: '16%',
-    padding: '2%'
+    padding: '2%',
+    pointerEvents: 'none'
   },
   pondPanelRight: {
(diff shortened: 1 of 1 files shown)


uponthesun:
Great improvements!

What do you think about the case when the info panel is up and the fish is already selected - with this version, clicking the fish again deselects it. Do you think this is better than just leaving it selected?

breville:
Good question.  Since we allow repeated clicks on a fish to toggle its selection in non-info mode, I think it's worth being consistent while in info mode.  Happy to revisit after we use it for a bit though.

PR #292, opened 2019-11-27
Remove X and tilda bodies

bethanyaconnor:


test/unit/qualityTest.test.js
```diff
@@ -200,14 +200,14 @@ describe('Model quality test', () => {
   // });
 
   test('test eels with sharp teeth', async () => {
-    const trainSize = 300; // Need more fish to hit enough to train on
+    const trainSize = 500; // Need more fish to hit enough to train on
```
bethanyaconnor: I had to increase this now that we only have one eel body...

uponthesun: That worries me - I thought we agreed to make each distinct "kind" of part more equally represented. Seems like we should get rid of both eels or neither.

bethanyaconnor:
Okay I'll update the test with a new test.

bethanyaconnor:
Closing in favor of #293

PR #296, opened 2019-11-27
More small tweaks

madelynkasula:
- Remove white outline on some buttons (follow-up: do this in Code Studio as well)
- Make "Play Again" / continue buttons show up on results screen sooner
- Temporarily disable battery asset
- Remove M from soda cup
- Add button to info guide:
Screen Shot 2019-11-27 at 10 06 51 AM


src/demo/ui.jsx
```diff
@@ -1505,6 +1504,11 @@ let Guide = class Guide extends React.Component {
                         : currentGuide.text}
                     </div>
                   </div>
+                  {currentGuide.style === 'Info' && (
+                    <Button style={styles.infoGuideButton} onClick={() => {}}>
```
breville: Does this button still work with an empty `onClick` handler?

madelynkasula: yep -- since any click (after typing has finished) closes the dialog, we got this for free 😂

PR #300, opened 2019-11-27
Revert "Revert "Factor fin color into model""

bethanyaconnor:
Reverts code-dot-org/ml-activities#299


uponthesun:
I merged from master to pick up change to increase Travis timeout.

PR #301, opened 2019-11-27
Only use the fish components for FishLong in the eels test

bethanyaconnor:
Most of this diff is from running prettier.


uponthesun:
I think the body shape test is failing now because the rest of the test code still loops through each body variation, not just ones found in short words mode, so there's a mismatch between the labelFns and the generated oceans.

uponthesun:
Also, if https://github.com/code-dot-org/ml-activities/pull/302 is merged first, can you have the colors test be in short words mode also? Thanks!

PR #302, opened 2019-11-27
Re-enable + fix color quality test

madelynkasula:


test/unit/qualityTest.test.js
```diff
@@ -174,30 +174,48 @@ describe('Model quality test', () => {
...
-  //     expect(result.recall).toBeGreaterThanOrEqual(0.5);
-  //   }
-  // });
+  test('Color test', async () => {
+    const trainSize = TRAIN_SIZE;
+
+    const idsByColor = {
+      red: [6, 7],
+      green: [2, 3],
+      blue: [4, 5]
```
madelynkasula: @uponthesun do you think i should add more colors here? downside: it makes the test take longer; upside: ?

test/unit/qualityTest.test.js
```diff
@@ -174,30 +174,48 @@ describe('Model quality test', () => {
...
+    };
+
+    const colorMatch = (rgb, colorIds) => {
+      const rgbs = Object.values(fishData.colors).map(color => color.rgb);
+
+      let match = false;
+      colorIds.forEach(id => {
```
uponthesun: I find the inner `return` statement from this block confusing, can we use the `for(const id of colorIds) {}` syntax instead so it just sets the `match` variable normally and then `break`s? Feel free to push back if you don't like that 😁

test/unit/qualityTest.test.js
uponthesun: I think this test should specifically be for the short words mode, so we only need to include the ones listed there. So I think this is fine.

test/unit/qualityTest.test.js
madelynkasula: ah, yeah, that would read better. i'll update!

PR #303, opened 2019-11-27
Fix bug in summarize on selecting maximum ID field per part

uponthesun:
The intent of this code to generate `idFieldsSummary` is to pick the maximum absolute value for each part type. The changed line is supposed to check to see if `idFieldsSummary` already contained that part key (e.g. `bodies`, `mouths`, etc.), and if not initialize it. However since previously I'd been calling `hasOwnProperty` on the wrong field, it never found the property, so the if statement would evaluate to true, with the end result that we were actually selecting the *minimum* value, not the max, since the list is sorted. Comedy of errors, really.

The impact of the bug was noticeable before when you trained on dorsal fins or tails - you would never see them as most important in the overall summary. After the fix they will show up there.

Anyway, the new `test SVM explanation fast fish` fails before the change and passes after to verify the fix.

I also added a dorsal fins test + reenabled assertions on accuracy for tails, since those were previously blind spots in the tests.

### DIFF
--- a/test/unit/qualityTest.test.js
+++ b/test/unit/qualityTest.test.js
@@ -305,6 +305,7 @@ describe('Model quality test', () => {
     }
   });
 
+
   test('test tails', async () => {
     const partData = fishData.tails;
     const partKey = PartKey.TAIL;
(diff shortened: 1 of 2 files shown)


test/unit/qualityTest.test.js
```diff
@@ -305,6 +305,7 @@ describe('Model quality test', () => {
     }
   });
 
+
```
bethanyaconnor: nit: stray extra line

test/unit/qualityTest.test.js
```diff
@@ -324,13 +325,37 @@ describe('Model quality test', () => {
...
+  test('test dorsal fins', async () => {
+    const partData = fishData.dorsalFins;
+    const partKey = PartKey.DORSAL_FIN;
+    const trainSize = TRAIN_SIZE;
+
+    for (const [name, data] of Object.entries(partData)) {
+      console.log(`${partKey} ${name}`);
+      const id = data.index;
+      const labelFn = fish => {
+        //console.log(JSON.stringify(fish, null, 2));
```
bethanyaconnor: nit: extra console log?

PR #307, opened 2019-11-28
fix keep training reminder

simplycloud:
fixing the reminder in the guide on part 1 after 15 objects trained; the Did You Know box appears, then the prompt that we want to include "Keep Training". Tested locally.

image

### DIFF
--- a/src/demo/models/guide.js
+++ b/src/demo/models/guide.js
@@ -80,7 +80,7 @@ const guides = [
   },
   {
     id: 'fishvtrash-training-pause4',
-    text: `A.I. learns from your choices.  If you make “wrong” choices, A.I. will learn to repeat your mistakes. Keep training.`,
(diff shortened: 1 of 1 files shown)


src/demo/models/guide.js
```diff
@@ -80,7 +80,7 @@ const guides = [
   },
   {
     id: 'fishvtrash-training-pause4',
-    text: `A.I. learns from your choices.  If you make “wrong” choices, A.I. will learn to repeat your mistakes.`,
+    text: `A.I. learns from your choices.  If you make “wrong” choices, A.I. will learn to repeat your mistakes. Keep training.`,
```
breville: super nit: double space after period to be consistent?

simplycloud: whoops, thanks for pointing that out.

PR #309, opened 2019-11-28
Fix recall button styling when info button is not displayed

madelynkasula:
bug Anthony found in Slack ([thread](https://codedotorg.slack.com/archives/CMV01LWPL/p1574921121054700?thread_ts=1574906734.048900&cid=CMV01LWPL))

Screen Shot 2019-11-27 at 10 56 47 PM

### DIFF
--- a/src/demo/ui.jsx
+++ b/src/demo/ui.jsx
@@ -128,7 +128,7 @@ const styles = {
   },
   confirmationDialogImg: {
     position: 'absolute',
-    bottom: '-47%',
+    bottom: '-46%',
     left: '-41%',
     height: '100%'
   },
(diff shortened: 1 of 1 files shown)


src/demo/ui.jsx
```diff
@@ -603,7 +603,7 @@ const styles = {
   },
   arrowUpperRight: {
     top: '13%',
-    right: '-3.5%',
+    right: '-2%',
```
breville: Actually do we know where the arrow should be shown, since the info button is never guaranteed?

breville: Oh, I think we only use arrow `UpperRight` on this early level so maybe nothing to worry about.

PR #312, opened 2019-12-01
Possible loading UI

breville:
A pairing exercise by @breville and @madelynkasula!

Adds a placeholder loading screen:
Screen Shot 2019-12-02 at 9 11 09 AM

In 2 places:
- `Modes.Loading`
- `Modes.IntermediateLoading` (a new mode; currently only used when going from training to sorting in part 2)

Will most likely wait to merge/ship this until we have the final asset from Mark, unless we decide to ship as-is and add the .gif from Mark later.


src/demo/models/loading.js
```diff
@@ -42,5 +44,17 @@ export const init = async () => {
...
+  const minimumEndTime = startTime + 2000;
+  const currentTime = $time();
+  let delayTime;
+  if (currentTime >= minimumEndTime) {
+    // We are already past the minimumEndTime, so don't wait any more.
+    delayTime = 0;
+  } else {
+    // We are at less than the minimumEndTime, so just wait the remainder of time.
+    delayTime = minimumEndTime - currentTime;
+  }
```
breville: I guess this can be simplified to:

```
  const delayTime = Math.max(0, minimumEndTime - $time());
  setTimeout(() => toMode(mode), delayTime);
```

though then I find I have to reverse-engineer the code, back to what's committed, to understand what it's doing.

madelynkasula:
@breville I refactored the approach for the Training -> Sorting loading screen -- can you take another look?

src/demo/models/loading.js
```diff
@@ -42,5 +44,7 @@ export const init = async () => {
   } else {
     mode = Modes.Words;
   }
-  toMode(mode);
+
+  // Ensure that we show the loading UI for at least 2 seconds, to avoid a flash.
```
breville: Can update duration.

breville:
@madelynkasula Looks good.  Let's manually test this a bunch.  The one aesthetic improvement I'd like is to get the initial loading screen to white; it currently pauses on grey because the UI thread is blocked mid-fade.

PR #313, opened 2019-12-01
improve prompts to encourage more training

simplycloud:
changes to a few prompts to emphasize connection between amount of training data and results.

1. change to part 2a to encourage the student to try continuing after only 5 training attempts to see the result. from the screen shots you can see the new guide prompt, and the results screen with lots of non circular fish after only training 5 times. we then encourage them to train more for a better result.
image
image
2. change to text around training more than 100 in part 2b and breaking prompts for train more or new word on the final screen to help students consider training more if they are not happy with the results screen.
image
image

### DIFF
--- a/src/demo/models/guide.js
+++ b/src/demo/models/guide.js
@@ -303,12 +303,12 @@ const guides = [
   },
   {
     id: 'fishshort-words-training-pause1',
-    text: `Nice work.  Keep training.`,
+    text: `Do you think A.I. has enough training data?  You could click Continue to find out.`,
     when: {
       appMode: AppMode.FishShort,
       currentMode: Modes.Training,
       fn: state => {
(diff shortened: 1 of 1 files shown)


src/demo/models/guide.js
```diff
@@ -303,12 +303,12 @@ const guides = [
   },
   {
     id: 'fishshort-words-training-pause1',
-    text: `Nice work.  Keep training.`,
+    text: `Do you think A.I. has enough training data?  You could click Continue to find out.`,
```
breville: We normally use "can" instead of "could", but perhaps this way it feels less like a suggestion, and more like an idea?

simplycloud: fixed, good catch!

PR #318, opened 2019-12-02
Loading gif pt. 2

madelynkasula:
Updating the "Loading..." component with a loading GIF:

Screen Shot 2019-12-02 at 1 16 33 PM

Also removed the overlays on loading screens.


src/demo/ui.jsx
```diff
@@ -751,7 +752,9 @@ let Loading = class Loading extends React.Component {
   render() {
     return (
       <Body>
-        <div style={styles.loading}>Loading...</div>
+        <div style={styles.loading}>
+          <img src={loadingGif} style={{maxWidth: 300}} />
```
breville: Should this be a %age for responsive displays?

madelynkasula: yeah, that sounds good. i updated it to 50% maxWidth

breville: Thanks.  To be finicky, I wonder if we should ensure that the loading bot is never larger than the actual bot shown in the tutorial?

madelynkasula: the bot sizing is done in different places, so i'm not sure we can 100% ensure that, but i see elsewhere that the bot is 30% so i'll set it to that instead

madelynkasula:
heads up: something is timing out in our Travis builds ([link](https://travis-ci.org/code-dot-org/ml-activities/branches) to the most current build on this PR), but i'm going to merge this PR because it needs to go out today. i verified locally that `yarn lint` and `yarn test` pass

PR #320, opened 2019-12-02
Increase Travis timeout from 10 to 20 min

madelynkasula:


madelynkasula:
closing this for 2 reasons: 1) this change didn't work, and 2) Travis is no longer timing out on builds

PR #327, opened 2019-12-03
Add maxWidth to info modal typing text

madelynkasula:
@breville noticed a width issue on the info modals' typing text:
Screen Shot 2019-12-03 at 1 12 46 AM

Fixed by setting a maxWidth on the typing text for "Info" guides only:
Screen Shot 2019-12-03 at 10 16 34 AM


madelynkasula:
i'm going to merge this so we can cut a release and get this into today's deploy

PR #330, opened 2019-12-04
More simple model tests

bethanyaconnor:
Again, just the beginning of adding some tests and trying to capture some our expectations around each model.


test/unit/demo/models/pond.test.js
```diff
@@ -0,0 +1,51 @@
+const {initFishData} = require('../../../../src/utils/fishData');
```
breville: Can we use some kind of root path to avoid all of these `..`?

test/unit/demo/models/predict.test.js
```diff
@@ -0,0 +1,43 @@
...
+    init();
+    // There's a setTimeout in the predict init. Ideally this would be
+    // refactored to use promises if possible but until then, wait 100 ms.
+    setTimeout(async () => {
+      const state = getState();
+      expect(state).toBeTruthy();
+      expect(state.mode).toBe(Modes.Predicting);
+      expect(state.fishData).toBeTruthy();
+      expect(state.fishData.length).toBeGreaterThan(0);
+    });
```
breville: Are we missing the `100` here?

bethanyaconnor: What do you mean?

bethanyaconnor: Discussed offline and fixed.

test/unit/demo/models/pond.test.js
madelynkasula: ➕ we should be able to add `alias`es to our webpack config to do this -- [example in code-dot-org](https://github.com/code-dot-org/code-dot-org/blob/988ae4a9e030d05eb917649ddcb6cadaf94406af/apps/webpack.js#L32-L92)

bethanyaconnor: Maddie did this in #332 and I've updated these tests (not any other tests though)

PR #331, opened 2019-12-04
Manually trigger GA events for IntermediateLoading and Predicting modes

madelynkasula:


src/demo/models/predict.js
```diff
@@ -27,6 +30,9 @@ export const init = () => {
   // the loading UI first.  If we are going to show the loading spinner, then also
   // delay the beginning of our training.
   setTimeout(() => {
+    // Manually send a GA event for Modes.Predicting.
+    setPageAndSendPageview('predicting');
```
breville: Oh, good catch!

src/demo/helpers.js
```diff
@@ -212,3 +212,13 @@ export const finishLoading = (startTime, onComplete) => {
 
   setTimeout(onComplete, delayTime);
 };
+
+export const setPageAndSendPageview = page => {
```
breville: I think it would be fine to leave "setPage" as an internal detail and just call this something like "reportPageView".

src/demo/models/index.js
```diff
@@ -27,20 +28,15 @@ export const init = state => {
...
-    window.ga('send', 'pageview');
+  const modeToPage = {
+    [Modes.Loading]: 'loading',
+    [Modes.Words]: 'words',
+    [Modes.Training]: 'training',
+    [Modes.Pond]: 'pond',
+    [Modes.Instructions]: 'instructions'
+  };
+  const page = modeToPage[state.currentMode];
+  if (window.ga && page) {
```
breville: I think it might be good to hoist all checks for `window.ga` inside the function rather than worrying about checking for it here.

bethanyaconnor:
Why do these need to be manually triggered?

madelynkasula:
@bethanyaconnor the code that writes to GA lives in `models/index.jsx` and is called every time we call `toMode()`:

https://github.com/code-dot-org/ml-activities/blob/981c9e7d4e29a1cb589b73ac20742ade08e80d8b/src/demo/models/index.js#L29-L45

because `Modes.IntermediateLoading` happens _during_ `Modes.Predicting`, we need to manually report those pageviews in `predict.js#init`

bethanyaconnor:
This is fine for now but I feel like it would be cleaner to just call `toMode` to transition to `Modes.IntermediateLoading`. Is there a reason we don't do that?

madelynkasula:
@bethanyaconnor yeah, good point. i didn't do that originally because it was more complex and i didn't want to introduce a regression, but i'll take a deeper look into that now

PR #332, opened 2019-12-05
Add @ml alias to /src/ dir

madelynkasula:
Now we can reference the `ml-activities/src/` dir as just `@ml/` to make our imports a little cleaner. I updated the `helpers.test.js` file in this PR to show an example, but didn't want to update every import in this PR.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,96 +1,109 @@
-const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+const {CleanWebpackPlugin} = require('clean-webpack-plugin');
 const CopyPlugin = require('copy-webpack-plugin');
+const path = require('path');
 
 const commonConfig = {
   devtool: 'eval-cheap-module-source-map',
   resolve: {
     extensions: ['.js', '.jsx'],
+    alias: {
+      '@ml': path.resolve(__dirname, 'src')
+    }
   },
   output: {
     filename: '[name].js',
-    libraryTarget: 'umd',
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -1,96 +1,109 @@
...
 const CopyPlugin = require('copy-webpack-plugin');
+const path = require('path');
 
 const commonConfig = {
   devtool: 'eval-cheap-module-source-map',
   resolve: {
     extensions: ['.js', '.jsx'],
+    alias: {
+      '@ml': path.resolve(__dirname, 'src')
+    }
```
madelynkasula: this is the only actual change in this file -- everything else (minus the new `require` statement on line 3) is prettier formatting

webpack.config.js
```diff
@@ -1,96 +1,109 @@
-const { CleanWebpackPlugin } = require('clean-webpack-plugin');
+const {CleanWebpackPlugin} = require('clean-webpack-plugin');
 const CopyPlugin = require('copy-webpack-plugin');
+const path = require('path');
 
 const commonConfig = {
   devtool: 'eval-cheap-module-source-map',
   resolve: {
     extensions: ['.js', '.jsx'],
+    alias: {
```
bethanyaconnor: Is it possible to add a comment that these need to be added to package.json if we want to use them in tests?

madelynkasula: yes, good call! added a note about this

PR #333, opened 2019-12-05
Don't render UI on every frame

breville:
While it appears to be a no-op in React and hopefully hasn't been harmful to our performance, we wanted to avoid telling React to update its UI every frame.  There were a few cases where we were still doing that, but didn't need to be.

### DIFF
--- a/src/demo/renderer.js
+++ b/src/demo/renderer.js
@@ -135,27 +135,39 @@ export const render = () => {
 
       if (state.appMode === AppMode.CreaturesVTrashDemo) {
         if (state.biasTextTime) {
-          setState({
-            canSkipPredict:
-              $time() >= state.biasTextTime + timeBeforeCanSkipBiasText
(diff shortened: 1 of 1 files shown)


src/demo/renderer.js
```diff
@@ -135,27 +135,39 @@ export const render = () => {
...
-        canSeePondText:
-          $time() >= currentModeStartTime + timeBeforeCanSeePondText
-      });
+      if (
+        !state.canSkipPond &&
+        $time() >= currentModeStartTime + timeBeforeCanSkipPond
+      ) {
+        setState({canSkipPond: true});
+      }
+      if (!state.canSeePondText) {
```
bethanyaconnor: This shouldn't be just a simple flip should it? There's a timeout here.

breville: Oops, you're right!

PR #336, opened 2019-12-09
Translation support for AI for Oceans

bethanyaconnor:
Adds `oceans.json` which will be where all source strings for the AI for Oceans activity will be added. This will then be exposed to anyone who has downloaded the package. We compile the English versions of these strings and then merge any strings we get from the initializer. This means that any missing strings (eg due to version differences) will simply default to English.

For the guide: The keys for the strings need to match the id for the guide.
For `ui.jsx`: the strings can just be called from the i18n object in state.

Adding the rest of strings can happen over the next few days.

https://github.com/code-dot-org/code-dot-org/pull/32362 implements this on the `code-dot-org` side.

### Full process for translating strings
In the `code-dot-org` repo, we use `MessageFormat` in our apps code. This repo will use the same version. The steps the strings will go through to be translated are:
1. Strings are added to `oceans.json`
2. The i18n pipeline will copy this file into a source directory and upload it to Crowdin
3. The strings get translated on Crowdin
4. The translations are downloaded and distributed to somewhere where the code that initializing the tutorial can access them.
5. The translations are passed into the initialization of the tutorial and stored in the `i18n` object in state.

### DIFF
--- a/bin/compileI18n.js
+++ b/bin/compileI18n.js
@@ -1,0 +1,7 @@
+var fs = require('fs');
+var MessageFormat = require('messageformat');
+
+const data = fs.readFileSync('i18n/oceans.json');
+const mf = new MessageFormat('en');
+const i18n = mf.precompileObject(JSON.parse(data));
+fs.writeFileSync('i18n/oceans.js', 'export const oceans_locale = ' + i18n);
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.nyc_output
 /.idea
 .DS_Store
+i18n/*.js
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@
 /.nyc_output
 /.idea
 .DS_Store
-i18n/*.js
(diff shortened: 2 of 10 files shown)


bin/compileI18n.js
```diff
@@ -0,0 +1,7 @@
+var fs = require('fs');
+var MessageFormat = require('messageformat');
+
+const data = fs.readFileSync('i18n/oceans.json');
+const mf = new MessageFormat('en');
+const i18n = mf.precompileObject(JSON.parse(data));
+fs.writeFileSync('i18n/oceans.js', 'export const oceans_locale = ' + i18n);
```
madelynkasula: should we prepend a comment when we write to `i18n/oceans.js` that explains that it's a generated file and shouldn't be edited manually?

breville:
Just curious: any notes on why this was closed?

bethanyaconnor:
I updated the approach and opened a new PR #343. As the only comment on this PR doesn't relate to the new PR, it was easier to just close this one and open a new one.

PR #340, opened 2019-12-11
More tests

bethanyaconnor:
Apparently none of the assertions we run previously so there was essentially no tests being run. This refactors the predict init function to return a promise, which the tests are much better at handling.

This diff looks a lot less scary if you hide whitespace changes. :)


test/unit/oceans/models/predict.test.js
```diff
@@ -1,45 +1,63 @@
 const {initFishData} = require('@ml/utils/fishData');
 import {setState, getState, resetState} from '@ml/oceans/state';
-import {ClassType, Modes} from '@ml/oceans/constants';
+import {ClassType, Modes, AppMode} from '@ml/oceans/constants';
 import {init, predictFish} from '@ml/oceans/models/predict';
 import SimpleTrainer from '@ml/utils/SimpleTrainer';
+import {TrashOceanObject} from '@ml/oceans/OceanObject';
 
-describe('Model quality test', () => {
+global.window = {};
```
madelynkasula: you should be able to remove this line now that `global.window` is set up in our jest config: https://github.com/code-dot-org/ml-activities/blob/3c05b9b14a358977ae9bf26591dbd7e2849ac047/package.json#L18-L19

bethanyaconnor: Thanks for catching! Removed.

PR #342, opened 2019-12-12
Unit test <Train/> component

madelynkasula:


test/unit/oceans/ui.test.js
```diff
@@ -197,3 +198,124 @@ describe('Words', () => {
...
+    expect(initialState.showConfirmationDialog).toBeFalsy();
+    expect(initialState.confirmationDialogOnYes).toBeNull();
+
+    const wrapper = shallow(<Train {...DEFAULT_PROPS} />);
+    const eraseIcon = wrapper.find('FontAwesomeIcon').at(0);
+    eraseIcon.simulate('click');
+
+    const newState = getState();
+    expect(newState.showConfirmationDialog).toBeTruthy();
+    expect(newState.confirmationDialogOnYes).not.toBeNull();
```
breville: Should we also test that the dialog has actually shown via the React DOM?

madelynkasula: the actual rendering happens in [`<UI/>`](https://github.com/code-dot-org/ml-activities/blob/e205ba00e4ff41ff37c7611fe041135f83c1ea22/src/oceans/ui.jsx#L1611-L1616), so i'll test that when i get there!

PR #343, opened 2019-12-12
Translating AI for Oceans

bethanyaconnor:
I created an `I18n` wrapper for messageformat. `I18n.t` isn't particularly necessary but I liked having the init code separated out.

The big downside to this approach currently is that it compiles the English strings on page load. I did some profiling and it takes 7ms on my machine (compared to over 100ms to initialize sounds) and will occur in the `Loading` state, so I think this is fairly acceptable.

The corresponding code-dot-org PR is https://github.com/code-dot-org/code-dot-org/pull/32362


src/oceans/models/guide.js
```diff
@@ -49,12 +59,15 @@ const guides = [
       }
     },
     style: 'Info',
+    didYouKnow: true,
```
madelynkasula: i'm not sure we need this extra property on guides -- `guide.style === 'Info'` should work in place of `guide.didYouKnow`

(cc @breville to double-check this thought)

src/oceans/ui.jsx
```diff
@@ -37,6 +31,7 @@ import {
   faInfo,
   faTrash
 } from '@fortawesome/free-solid-svg-icons';
+import * as I18n from './i18n';
```
madelynkasula: will unit tests break if we don't call `I18n.init()` before we test a component that uses i18n strings? if they do, we should make this module mockable so that we aren't depending on it being initialized before every test

i18n/oceans.json
```diff
@@ -0,0 +1,107 @@
...
+    "creaturesvtrash-training-init5": "Keep training to teach A.I. what belongs in the water.",
+    "creaturesvtrash-training-init6": "Great work!  The more data you provide A.I., the better it will do.  Keep training A.I. or continue when ready.",
+    "creaturesvtrash-predicting-init1": "Do you think A.I. will now do a better job identifying what should be in the water?  Let’s watch.",
+    "creaturesvtrash-pond-init1": "Based on your training, here are some objects that A.I. identified as “belongs in water”.  How did A.I. do?",
+    "fishshort-words-init1": "Artificial intelligence and machine learning can be used to teach a computer new patterns.",
+    "fishshort-words-init2": "Let’s teach A.I. a new word by showing it examples of that type of fish.",
+    "fishshort-words-training-pause1": "Do you think A.I. has enough training data?  You can click Continue to find out.",
+    "fishshort-words-training-pause2": "Great work!  Keep training A.I. or continue when ready.",
+    "fishshort-predicting-init1": "With your training data, you have programmed A.I. to recognize “{WORD}” fish.",
```
madelynkasula: do the `WORD` and `N` keys need to be all caps? i think our/js convention is camelcase for i18n args

bethanyaconnor: Don't think so. I was following some documentation but if our convention is camel-case I'll change it.

src/oceans/ui.jsx
bethanyaconnor: hm I'm not sure but our unit tests are passing

bethanyaconnor: oh lol nevermind they're not now that I've merged master in. I'll definitely fix the failing test cases

madelynkasula: we talked offline, but commenting for future us:

i lied, we should actually just call `I18n.init()` in our tests because we have some unit tests that depend on accurate translations:

https://github.com/code-dot-org/ml-activities/blob/e144c663de0592cbcf3a312870d3ae6833bb0961/test/unit/oceans/ui.test.js#L270

and it would be more annoying to mock every return value from i18n than it is to just initialize it at the beginning of test suites that need it

src/oceans/models/guide.js
bethanyaconnor: I'm okay with that. I've made this change -- let me know if there are any objections.

package.json
```diff
@@ -101,6 +101,7 @@
     "@fortawesome/fontawesome-svg-core": "^1.2.25",
     "@fortawesome/free-solid-svg-icons": "^5.11.2",
     "@fortawesome/react-fontawesome": "^0.1.7",
+    "messageformat": "^1.0.0",
```
Hamms: is there a specific reason to use v1.0 rather than the latest? (v2.3)

src/oceans/i18n.js
```diff
@@ -0,0 +1,12 @@
+import data from '../../i18n/oceans.json';
+import MessageFormat from 'messageformat'
+
+let messages;
+export const init = (i18n = {}) => {
+  const mf = new MessageFormat('en');
```
Hamms: I don't love that we're hardcoding 'en' here even when we're actually rendering in another language; but I also don't see that this matters enough for us to go through all the work of passing locale in from code-dot-org.

Could you maybe just add a comment here explaining why we're doing what we're doing?

src/oceans/init.js
```diff
@@ -7,6 +7,7 @@ import {setInitialState, setSetStateCallback} from './state';
 import {render as renderCanvas} from './renderer';
 import modeHelpers from './modeHelpers';
 import soundLibrary from './models/soundLibrary';
+import * as I18n from './i18n';
```
Hamms: any reason not to just build this as a class if we're going to treat it as a class?

src/oceans/i18n.js
```diff
@@ -0,0 +1,12 @@
...
+  messages = {...mf.compile(data), ...i18n};
+};
+
+export const t = (key, options) => {
+    return messages[key](options);
```
Hamms: it'd be nice to have something here like
```suggestion
    if (!messages) {
      throw "I18n must be initialized before it can render translations"
    }
    return messages[key](options);
```

package.json
bethanyaconnor: 2.3 requires babel 7 (I tried to upgrade this repo to use babel 7 and hit a number of road blocks). I get this error:  ```ReferenceError: Unknown plugin "@babel/plugin-proposal-class-properties" specified in "/home/bconnor11/ml-activities/node_modules/messageformat/package.json" at 0, attempted to resolve relative to "/home/bconnor11/ml-activities/node_modules/messageformat"```

I can use v1.1 though so I'll upgrade to that.

src/oceans/init.js
bethanyaconnor: Changed this a bit to be consistent with most of our other files (like `soundLibrary`).

PR #344, opened 2019-12-16
<Predict/> tests

madelynkasula:


src/oceans/ui.jsx
```diff
@@ -1040,14 +1040,12 @@ let Predict = class Predict extends React.Component {
       this.state.timeScale !== defaultTimeScale
     ) {
       selectedControl = MediaControl.FastForward;
-    } else {
-      selectedControl = MediaControl.Play;
```
madelynkasula: when we set `selectedControl = MediaControl.Play`, it was never being used, so i just removed this case

src/oceans/ui.jsx
```diff
@@ -1040,14 +1040,12 @@ let Predict = class Predict extends React.Component {
...
     }
 
     return (
       <Body>
         {this.state.displayControls && (
-          <div style={styles.mediaControls}>
+          <div style={styles.mediaControls} id="uitest-media-ctrl">
```
breville: Is there any way to avoid using DOM IDs, say by using React references to these elements?  Or is this the preferred technique?

madelynkasula: i was following the convention in the code-dot-org repo, which is to use IDs prefaced with 'uitest,' so i think this is the preferred technique

PR #345, opened 2019-12-17
Unit test <Pond/>

madelynkasula:


src/oceans/ui.jsx
```diff
@@ -1364,13 +1362,10 @@ let Pond = class Pond extends React.Component {
     }
   };
 
-  onPondPanelButtonClick(e) {
+  onPondPanelButtonClick = e => {
```
madelynkasula: @breville i realized that this is why you were seeing weirdness when trying to use `onClick={this.onPondPanelButtonClick}` (instead of `onClick={e => this.onPondPanelButtonClick(e)}`) -- only "fat-arrow"/ES6 functions are automatically bound ([stackoverflow](https://stackoverflow.com/questions/46493774/es5-this-method-is-not-a-function))

breville: Oh, that makes sense (in hindsight)!  Thanks for remembering and noticing!

PR #350, opened 2019-12-20
Rename SimpleTrainer to KNNTrainer as that is a more descriptive name

bethanyaconnor:


bethanyaconnor:
Yeah it was a simple find/replace. I would hope that we have enough unit tests now that would catch an issue here, especially combined with a grep for `SimpleTrainer`.

Good idea on `knnData`! Maybe I'll take a stab if I ever get time but I think that cleanup should be in a separate PR

PR #352, opened 2019-12-20
Start of RTL support

bethanyaconnor:
This just sets the direction of the bot (and its head) to be `ltr` all the time as well as the recall buttons.

There's still some weirdness with button padding in RTL on the local build but I couldn't repro them on code-dot-org. The buttons are flipped in RTL, which is probably okay, but I'm open to feedback. 

Code studio screenshot (with a couple translations I pulled off Crowdin):
Screenshot 2019-12-20 at 2 07 06 PM

### DIFF
--- a/src/oceans/ui.jsx
+++ b/src/oceans/ui.jsx
@@ -236,22 +236,26 @@ const styles = {
     position: 'absolute',
     top: '30%',
     right: '-2%',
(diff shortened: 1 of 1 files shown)


bethanyaconnor:
@madelynkasula verbally approved this PR to unblock cutting a new release.

PR #354, opened 2020-01-21
Add markdown support for strings

bethanyaconnor:
This PR adds the ability for strings to be written in Markdown to use custom styling. For this particular string, the `strong` and `em` tags are used and then overridden for the styling.

Because these styles are font-related, we will need to add them to the `code-dot-org` repo manually as well when we bump the version.


i18n/oceans.json
```diff
@@ -63,8 +63,7 @@
     "run": "Run",
     "mostImportantParts": "These were the most important fish parts:",
     "clickIndividualFish": "Click individual fish to see their information.",
-    "mostImportantPartsInDetermining": "These were the most important fish parts in determining whether this fish was",
-    "or": "or",
+    "mostImportantPartsInDeterminingFullSentence": "These were the most important fish parts in determining whether this fish was “*{word}*” or “**{notWord}**”.",
```
breville: `..."FullSentence"` seems a bit awkward.  How about just `"mostImportantParts"`?

src/oceans/ui.jsx
```diff
@@ -1180,18 +1181,18 @@ class PondPanel extends React.Component {
...
-                  {'”'}.
+                <div style={styles.pondPanelPreText} id="pondTextMarkdown">
+                  <Markdown
+                    markdown={I18n.t(
+                      'mostImportantPartsInDeterminingFullSentence',
+                      {
+                        word: state.word.toLowerCase(),
+                        notWord: I18n.t('notWord', {
+                          word: state.word
+                        }).toLowerCase()
```
breville: Can we simplify to just substitute in `state.word` and put the negation translation into the sentence itself?

bethanyaconnor: I went back and forth on this and decided to keep it this way because the sentence is pointing out actual strings in the UI (like the buttons).

i18n/oceans.json
bethanyaconnor: `mostImportantParts` already exists as a key :)

We also can't reuse the old keys (or else I would've used `mostImportantPartsInDetermining` again). The key isn't actually that important and I can't really think of anything better but definitely open to suggestions!

src/oceans/ui.jsx
breville: Oh, that's a good point!

i18n/oceans.json
Hamms: `mostImportantPartsDescription`?

bethanyaconnor:
@Hamms a quick attempt seems to strip out the HTML tags out of Markdown strings and just renders the HTML as plain text for the rest of them.

PR #378, opened 2021-09-08
STAR-1608: Fix missing dependency. Use @code-dot-org fork of svm

aaronwaggener:
One of our forked dependencies was deleted which was causing our builds to break. I recreated the dependency in https://github.com/code-dot-org/svmjs/pull/1 and have updated the references here.

[STAR-1608](https://codedotorg.atlassian.net/browse/STAR-1608)

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "remark-rehype": "^5.0.0",
     "sinon": "^7.5.0",
     "style-loader": "^1.0.0",
-    "svm": "uponthesun/svmjs",
+    "@code-dot-org/svm": "^0.1.1",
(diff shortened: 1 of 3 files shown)


madelynkasula (approved):
this change looks good. i think next we'll need to move this repo over to using github actions (it was using travis ci), right?

aaronwaggener:
> this change looks good. i think next we'll need to move this repo over to using github actions (it was using travis ci), right?

@madelynkasula, Yes, that's exactly right. First step was to get these running again. Next I will migrate to Actions so we can see 🟢 green builds 🟢 again.

PR #381, opened 2021-09-11
Migrate to GitHub Actions

aaronwaggener:
Follow-up from https://github.com/code-dot-org/ml-activities/pull/378

Replaces TravisCI with Github Actions. I created a Build action for PRs and a Deploy one for merges. The Deploy action essentially runs through the same build steps and then has an extra "action" to deploy to [GitHub Pages](https://code-dot-org.github.io/ml-activities/).

I also went ahead and moved this repo to node v14 to match the code-dot-org app. And I had to bump the `canvas` package up to one with an available linux binary (downloads during `yarn install`).


.github/workflows/build.yml
```diff
@@ -0,0 +1,22 @@
+name: Build
```
madelynkasula, line 1: in the maker app, the filename for the "Build" action is [main.yml](https://github.com/code-dot-org/browser/blob/main/.github/workflows/main.yml). i think they should be the same, but i'm not sure which convention makes more sense. maybe this one (build.yml after the job rather than the branch)?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '14.17.1'
+      
+      - name: Install and Build
+        run: |
+          yarn install --frozen-lockfile
+          yarn test
+          yarn build
```
madelynkasula, line 24: could we have this action depend on the build action to remove the duplication here? it looks like we could use the [`needs` parameter](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds). all examples i'm seeing have the 2 jobs in the same file, so i'm not sure if we'll run into issues having them in separate files

.github/workflows/build.yml
aaronwaggener: Given that we now have a case where we have multiple jobs in one repo, I think it makes more sense to have these file names match the job names. I will update the Maker App job.

.github/workflows/deploy.yml
aaronwaggener: Nice find! This duplication did not feel quite right. I will see if I can get this working with `needs`. 

If we do have to keep the jobs together in a single file, that might affect where we land on the naming convention (per the earlier comment).

.github/workflows/build.yml
aaronwaggener: After a little research I think ci.yml with a descriptive "name" value is the right approach here.

.github/workflows/deploy.yml
aaronwaggener: Turns out `needs` does need to run within the same workflow (file). I could have created a separate job in the same file and used `needs` in there, but the artifacts from the `build` job would not have been available to it. There are ways to zip up and save artifacts and then pull them down in another job, but it felt like overkill for this use case so I added a simple `if` condition after the build step. If we want to do a more complex workflow in the future we can always break it out to a separate job.

aaronwaggener:
@madelynkasula ready for another review. Let me know if you think we should go a different direction with the naming and/or conditional deploy step.

madelynkasula (approved):
this looks great! i think the decision to keep build and deploy in a single job makes sense.

separate thought: most of our island repos don't deploy on merge (we manually publish an NPM package) -- i wonder if we should move toward having those repos automatically publish when we merge into `main` 🤔

aaronwaggener:
> most of our island repos don't deploy on merge (we manually publish an NPM package) -- i wonder if we should move toward having those repos automatically publish when we merge into `main` 🤔

@madelynkasula I was thinking about something along the same lines. The flow I had in mind was to still have a developer use `npm version` and the script in package.json would be a simple git tag and push. Then we have a workflow that will get triggered when a new tag is seen and it would kick off a build and publish from `main`. By triggering on tag there is still some intentionality from a developer to publish a new version and for them to choose what version we are bumping to. I would love to explore this further next time we have changes to a package we are publishing.

PR #401, opened 2024-06-27
Add regular suggestions for students to move on after training

bencodeorg:
Adds additional "guide" (ie, the bits of text that appear on screen in AI for Oceans) entries suggesting that students move on after training excessive amounts of data. Implements this in each of the three levels where training is occurring.

Also removes the first "Did you know?" tidbit that we pop up, as curriculum noted it to be disruptive to teachers introducing training to their students. This could easily be moved to a different point in the progression based on discussion here.


bencodeorg:
Closing in favor of https://github.com/code-dot-org/ml-activities/pull/406

PR #402, opened 2024-06-29
Guides variant: K5

breville:
With this change, AI for Oceans supports a new variant of guides for some new K-5 curriculum.  This means that the tutorial can still run unchanged from its original form as an Hour of Code tutorial, but the curriculum team can author a new set of guide messages, for the same modes as before, for new K-5 curriculum.

For now, this can be authored by running locally with URLs like the following:
- http://localhost:8080 runs the original Hour of Code guides.
- http://localhost:8080/?guides=K5 runs with the new K5 guides.

The original Hour of Code guides are now in `guidesHoc.js`, and should remain unchanged for now.  The new guides duplicate this set, with one example text change, in `guidesK5.js`, which can now be modified by the curriculum team.


src/oceans/models/guide.js
```diff
@@ -479,7 +10,9 @@ const getCurrentGuide = () => {
 
   const state = getState();
 
-  for (const guide of guides) {
+  const currentGuides = queryStrFor('guides') === 'K5' ? guidesK5 : guidesHoc;
```
bethanyaconnor, line 13: I assume this will be configurable on the studio side?

bethanyaconnor (commented):
What is the imagined way for curriculum writers to update this. Strings really need to be added to https://github.com/code-dot-org/ml-activities/blob/main/i18n/oceans.json, then use the `I18n.t` function to call them. Then we'll need to export this package. I assume this will be a one-time update as a result?

src/oceans/models/guide.js
breville: Great question.  I just updated this PR to make this a parameter for `initAll`, and then we will be able to configure this on the studio side.

ebeastlake:
Acknowledged/seems reasonable, pending approval by @bethanyaconnor. Thanks, @breville!

bencodeorg (approved):
LGTM -- just so we don't lose it, what's the "small change" you made to the K5 guides?

breville:
> LGTM -- just so we don't lose it, what's the "small change" you made to the K5 guides?

Oh, it's the test string [here](https://github.com/code-dot-org/ml-activities/pull/402/files#diff-c116fd73dcc5534b131136b95ae2506cc60b370f8bca5103e679d90a4742a140R154).

src/oceans/models/guidesK5.js
```diff
@@ -0,0 +1,485 @@
...
+  },
+  {
+    id: "fishvtrash-pond-init3",
+    textFn: () => I18n.t("fishvtrash-pond-init3"),
+    when: { appMode: AppMode.FishVTrash, currentMode: Modes.Pond },
+    arrow: "LowerRight",
+  },
+  {
+    id: "creaturesvtrashdemo-predicting-init1",
+    textFn: () => 'This is a test string.',
```
breville, line 154: Here is an example unlocalized string that can be used to confirm which variant is showing.

PR #403, opened 2024-07-01
Initial text to speech support

breville:
This adds initial text to speech support to AI for Oceans, intended for internal evaluation for starters.

It uses the browser's built-in [`SpeechSynthesis`](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis) API, which looks to now be widely-available.

A variety of designs were attempted in building this PR.  Currently, the user experience is that, if text to speech is enabled, typing sounds are not played, but the typing animation is still shown.  The completion of the typing is still used to determine when the guide can be dismissed (partly due to the observation of a bug in at least one language when the full text was not being read out, meaning that relying on the speech completing was not a reliable way to decide that the guide could now be dismissed).  

At least on Google Chrome on MacOS, playing the speech must be played in response to a user action.  If the user clicks anywhere while the guide is typing, and before we've first played speech, then we attempt to play the speech.  If the user clicks anywhere after the guide is done typing, and still before we've first played speech, then we make one attempt to play the speech; if the speech plays, then the next click will dismiss the guide.  Once speech has played once, then on subsequent displays of the guide we attempt to play speech immediately.  Speech is stopped whenever the guide is dismissed.  Note that speech often isn't available right away, as the browser appears to do an asynchronous load of the voices.  

We use the helpful https://github.com/HadrienGardeur/web-speech-recommended-voices in determining which voice to play.  Currently, we consume their JSON for English and Italian, and use the first available match.  Their demo at https://hadriengardeur.github.io/web-speech-recommended-voices/demo/ is helpful in hearing the variety of recommended voices for each language.

In this initial work, the standalone text to speech can be heard in English by loading with the appropriate URL parameter:
- http://localhost:8080/?tts=en: English text with the first English voice
- http://localhost:8080/?tts=it: English text with the first Italian voice

https://github.com/code-dot-org/ml-activities/assets/2205926/e5debc47-2f74-4700-9c9d-e8e1385310c5


src/oceans/components/common/Guide.jsx
```diff
@@ -1,29 +1,82 @@
...
+  stopTextToSpeech,
+  hasTextToSpeechVoices
+} from '@ml/utils/TextToSpeech';
+
+// A timer used for playing typing sounds.
+let guideTypingTimer = undefined;
+
+// Whether text to speech has ever been successfully
+// started via a user click.
+let textToSpeechStartedViaClick = false;
```
ebeastlake, line 24: Would `hasTextToSpeechStarted` or `hasTextToSpeechStartedByClick` be an easier-to-interpret variable name here?

src/oceans/models/guide.js
```diff
@@ -136,19 +136,37 @@ const guides = [
   {
     id: 'fishvtrash-pond-recall',
     textFn: () => I18n.t('fishvtrash-pond-recall'),
-    when: {appMode: AppMode.FishVTrash, currentMode: Modes.Pond},
+    when: {
+      appMode: AppMode.FishVTrash,
+      currentMode: Modes.Pond,
+      fn: state => {
```
ebeastlake, line 142: Why did these get added?

src/oceans/components/common/Guide.jsx
```diff
@@ -1,29 +1,82 @@
...
+
+    let textToSpeechStarted = false;
+
+    // Start playing text to speech.
+    if (
+      state.textToSpeechLocale &&
+      hasTextToSpeechVoices() &&
+      !textToSpeechStartedViaClick &&
+      textToSpeechCurrentGuide !== currentGuide &&
+      currentGuide
```
ebeastlake, line 54: Can we extract this into a helper function for readability, something like:
```
const shouldStartTextToSpeech = (state) => (
   state.textToSpeechLocale &&
   hasTextToSpeechVoices() &&
   !hasTextToSpeechStartedByClick/!textToSpeechStartedViaClick &&
   textToSpeechCurrentGuide !== currentGuide &&
   currentGuide
);
```

src/oceans/components/common/Guide.jsx
```diff
@@ -1,29 +1,82 @@
...
+        )
+      ) {
+        textToSpeechCurrentGuide = currentGuide;
+        textToSpeechStartedViaClick = true;
+        textToSpeechStarted = true;
+      }
+    }
+
+    if (!textToSpeechStarted) {
+      const dismissed = guide.dismissCurrentGuide();
```
ebeastlake, line 70: nit: I think a couple of comments could make this clearer. Maybe one here that says something like `Handle guide dismissal if we were unable to initiate text-to-speech (due to missing voices, etc.)` and one on line 63 to the effect of `Registers text-to-speech settings for the current guide if successful`.

src/utils/TextToSpeech.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  const filteredVoices = filterAvailableVoices(recommendedVoices[locale]);
+  if (filteredVoices.length === 0) {
+    return false;
+  }
+
+  const utterance = new SpeechSynthesisUtterance(text);
+
+  for (const voice of voices) {
```
ebeastlake, line 26: Potential follow-up: Should we store a list of `matchedVoicesByLocale` so we don't have to search the entire list every time?

ebeastlake (commented):
Nothing major, nice work @breville! It'll be good to get a second set of eyes.

src/oceans/components/common/Guide.jsx
```diff
@@ -1,29 +1,82 @@
...
+import {Button} from '@ml/oceans/components/common';
+import arrowDownImage from '@public/images/arrow-down.png';
+import {
+  startTextToSpeech,
+  stopTextToSpeech,
+  hasTextToSpeechVoices
+} from '@ml/utils/TextToSpeech';
+
+// A timer used for playing typing sounds.
+let guideTypingTimer = undefined;
```
bencodeorg, line 20: Is this no longer needed as state? I still see it referenced in a couple places (eg, `index.jsx`)

bencodeorg (commented):
Looks like Emily had a few notes as well, so putting these comments out there before it gets to be too much. I pulled locally though and this looks great overall!

src/utils/TextToSpeech.js
```diff
@@ -0,0 +1,74 @@
+const supportedVoiceLanguages = ['en', 'it'];
+let recommendedVoices = [];
```
bethanyaconnor, line 2: nit: I think you can use `const` over `let` here?

bethanyaconnor: I often get mixed up on what qualifies as an assignment for arrays in JS, though, so could be wrong!

bethanyaconnor: Also, should this be an object over an array? (sorry for the extensive comments at this point)

src/oceans/components/common/Guide.jsx
breville: Updated to `hasTextToSpeechStartedByClick`.

src/oceans/models/guide.js
breville: I noticed that we were temporarily entering some of these states prematurely because the condition wasn't enumerated completely.  While it wasn't noticeable with a text-rendered guide, it did mean we set up to speak a phrase that would be immediately stopped again.  Tidying this up made debugging easier, too.

src/utils/textToSpeech/README.md
```diff
@@ -0,0 +1 @@
+Files in here come from https://github.com/HadrienGardeur/web-speech-recommended-voices
```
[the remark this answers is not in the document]
breville: Nice find.  This would be great!

src/oceans/index.jsx
```diff
@@ -26,6 +27,7 @@ function initDemoPage() {
 
   initAll({
     appMode: currentAppMode,
+    textToSpeechLocale,
```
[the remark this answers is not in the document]
breville: This parameter can be sent from the dashboard side [here](https://github.com/code-dot-org/code-dot-org/blob/a042f0c1587e90d05b5bbdf35f17fd0a8a8fada8/apps/src/fish/Fish.js#L131), and yes, we can prep a change in the main repo to pass it using whichever criteria we decide upon.

src/oceans/components/common/Guide.jsx
breville: Calling `setState` from our `render` function was trigging a (justifiable) warning from React:
```
Render methods should be a pure function of props and state; triggering nested component updates from render is not allowed. If necessary, trigger nested updates in componentDidUpdate.
```

At first I hoisted some variables out of state to avoid this, since changes to them don't need to trigger re-renders, but then I remembered that we had an optional `skipCallback` parameter in `setState` so I used that instead.

src/utils/TextToSpeech.js
```diff
@@ -0,0 +1,74 @@
...
+    }
+  }
+
+  utterance.onend = onComplete;
+
+  speechSynthesis.cancel();
+
+  speechSynthesis.speak(utterance);
+
+  console.log('Saying:', text);
```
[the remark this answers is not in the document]
breville: Done.

src/oceans/index.jsx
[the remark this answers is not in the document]
bencodeorg: Ah I'm sorry, looking back at the K5 guides PR I totally misunderstood how you set up being able to configure via dashboard. LGTM!

src/oceans/components/common/Guide.jsx
breville: I refactored this code quite a bit, with a centralized `updateTextToSpeech` function that has a bunch of comments.  Hopefully that helps.

src/utils/TextToSpeech.js
breville: Oh, good calls.  Fixed.

src/utils/TextToSpeech.js
breville: This does make sense longer term.  One reason I have it do the whole search each time now, rather than caching, is that it seems possible that the browser's voice list will refresh more than once as more voices are loaded; the documentation for the [SpeechSynthesis voiceschanged](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis/voiceschanged_event) event isn't clear on whether this is likely to happen only once or more than once.  By reexamining the list of voices, it will give us the opportunity to hear if a better voice arrives partway through usage.

src/utils/TextToSpeech.js
```diff
@@ -0,0 +1,74 @@
...
+// From https://github.com/HadrienGardeur/web-speech-recommended-voices.
+function filterAvailableVoices(jsonData) {
+  if (!jsonData) {
+    return [];
+  }
+
+  const availableVoices = [];
+  const voices = speechSynthesis.getVoices();
+
+  jsonData.voices.forEach(function(voice) {
```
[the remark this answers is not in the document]
breville: Good question.  The voice list should already be ordered.  Excerpts from [here](https://github.com/HadrienGardeur/web-speech-recommended-voices?tab=readme-ov-file#guiding-principles):

> - Each voice list is ordered and meant to provide an optimal listening experience on all browsers/OS/languages covered by this project.
> - But each list also includes default options, to make sure that there's always something reliable to lean on.
> - With these two goals in mind, higher quality voices are listed on top of the list, while lower quality voices or specialized ones are listed at the bottom.

It would be great if they could produce an (opinionated) API as we [discussed](https://github.com/code-dot-org/ml-activities/pull/403/files#r1662992090).  

It might be worth borrowing more from their [demo code](https://github.com/HadrienGardeur/web-speech-recommended-voices/blob/2bbf8705e68d82bc2aea2395079c24f3f4a69dcd/demo/script.js), if we like its final voice sort better.  I'm not quite sure why, yet, but on macOS Chrome their additional sorting puts `Female Google voice (US)` (quality: `high`) at the top of the list, but our subset of the code puts `Daniel (UK)` (quality: `normal`) at the top.  `Daniel (UK)` is higher in the JSON file, possibly because of the potential 14 second playback bug with `Female Google voice (US)`, which I believe I have also observed, so it doesn't seem unreasonable.  Definitely worth further thinking, though.

src/utils/TextToSpeech.js
```diff
@@ -0,0 +1,74 @@
...
+          if (voices.some(apiVoice => apiVoice.name === altName)) {
+            voice.name = altName;
+            availableVoices.push(voice);
+          }
+        });
+      }
+    }
+  });
+
+  return availableVoices;
```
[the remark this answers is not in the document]
breville: I was thinking there might be some value in returning their array of results, in case we want to offer something like a voice selection UI, or allow the calling code to use additional selection heuristics.

Agreed on the confusion.  And also concerned by the way `voice` gets modified in the `altNames` case!  This code was borrowed direct from their demo, partly with the thought that improvements made there could be easily ported over.

src/oceans/components/common/Guide.jsx
breville: Hopefully the refactor helped address this.

src/oceans/components/common/Guide.jsx
```diff
@@ -1,28 +1,98 @@
...
+        soundLibrary.playSound('other');
+      }
     }
-  }
+  };
+
+  // Called from both the guide click handler and the render method, and
+  // attempts to play text to speech if needed.  Returns true if it believes
+  // it started text to speech.
+  updateTextToSpeech = inClickHandler => {
```
bencodeorg, line 62: Maybe "attemptTextToSpeech"?

src/oceans/components/common/Guide.jsx
```diff
@@ -41,11 +111,21 @@ let UnwrappedGuide = class Guide extends React.Component {
     }
 
     // Start playing the typing sounds.
-    if (!state.guideShowing && !state.guideTypingTimer && currentGuide) {
+    if (
+      !state.textToSpeechLocale &&
```
bencodeorg, line 115: Probably fine, but we'll be turning off typing sounds even if we aren't able to successfully play TTS for a given locale (if one is provided), yeah?

src/utils/TextToSpeech.js
bencodeorg: ...is this fixed?

src/utils/TextToSpeech.js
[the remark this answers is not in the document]
bencodeorg: Makes sense, thanks for the info!

src/utils/TextToSpeech.js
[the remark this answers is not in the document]
bencodeorg: Ah got it, good to know it's copied over. Good thought re: there maybe being value in returning an array of results, hadn't thought of that!

src/oceans/models/guidesK5.js
```diff
@@ -14,472 +14,496 @@ const imageStyleOverrides = {
 
 const guidesK5 = [
   {
-    id: "fishvtrash-training-init1",
+    id: 'fishvtrash-training-init1',
```
bencodeorg, line 17: These are just linting changes, yeah?

src/utils/TextToSpeech.js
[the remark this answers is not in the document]
bencodeorg: Looking at these a bit more, distinguishing `browserVoices` vs. `voicesMetadata` or something along those lines in var names might be helpful for future readers.

src/oceans/components/common/Guide.jsx
```diff
@@ -1,28 +1,98 @@
...
+    if (
+      !currentGuide ||
+      state.textToSpeechCurrentGuide === currentGuide
+    ) {
+      return false;
+    }
+
+    // If we are in a click handler, do nothing if we've already started
+    // text to speech from a click handler.
+    // If we are not in a click handler, do nothing if we've never started
```
bencodeorg, line 83: Took me a minute to follow here -- I think this bit I was missing in understanding the second case is that we always want to initiate TTS for the first time via click (right)? Could we update this comment to note that?

bencodeorg (approved):
This is going to be a huge improvement, nice work! Also wanted to call out the attention to detail around getting TTS to be initiated by a user action, making sure it terminates appropriately with dismissal of the guide, etc. A really nice experience!

src/oceans/models/guidesK5.js
breville: Yes.  For some reason my recent PR that updated these had the wrong linting.

src/oceans/components/common/Guide.jsx
breville: Yes.  It's definitely worth keeping an eye on whether there are situations where we'd prefer the sounds to be left on.

src/oceans/components/common/Guide.jsx
breville: Good call.  Updated.

src/utils/TextToSpeech.js
breville: Thanks for catching this; somehow the fix didn't get committed.  Should be updated now.

src/oceans/components/common/Guide.jsx
breville: Perfect.  Updated.

PR #406, opened 2024-07-11
K5 guides curriculum-specified updates

bencodeorg:
Makes changes to the "guides" presented in the new K5 version of this progression, which are generally:
- removing some "Did you know?" guides.
- add more reminders to suggest that users stop training once they've categorized a large number of fish/sea creatures.

The details of the changes requested are in [this doc](https://docs.google.com/document/d/1szxMX1deW-gBYJa3xigPDgVvgrMw3hEJ-scqr-bBigo/edit). Note that the `FishShort` app mode (which is used on [this level in the Hour of Code script](https://studio.code.org/s/oceans/lessons/1/levels/6)) is not used in the current K5 AI unit, so no changes are made there.

## Testing story

I ran through each of the affected levels and compared the guides I saw to the expected behavior on the doc linked above.

## Follow-up work

We'll need to hook this up to level config such that the right version of the guides file (`guidesK5`) is used in the K5 AI unit.

### DIFF
--- a/src/oceans/models/guidesK5.js
+++ b/src/oceans/models/guidesK5.js
@@ -12,6 +12,9 @@ const imageStyleOverrides = {
   can: {bottom: '2%', left: '16%'}
 };
 
+const encourageStopTrainingCountsFishLong = [150, 200, 250];
+const encourageStopTrainingCountsDefault = [100, ...encourageStopTrainingCountsFishLong];
+
 const guidesK5 = [
   {
     id: "fishvtrash-training-init1",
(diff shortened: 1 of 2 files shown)


src/oceans/models/guidesK5.js
```diff
@@ -99,10 +60,23 @@ const guidesK5 = [
...
+      textFn: () => I18n.t('training-generic-please-continue'),
+      when: {
+        appMode: AppMode.FishVTrash,
+        currentMode: Modes.Training,
+        fn: state => {
+          return state.yesCount + state.noCount >= count;
+        }
+      }
+    }
+  }),
```
breville, line 79: Nice approach!

breville, line 79: Nice approach!

breville (approved):

